PR #15, opened 2015-01-20
Changes corresponding to the change in Tango API spec.

mihirpandya:


restful-tango/tangoREST.py
```diff
@@ -137,9 +137,11 @@ def convertJobObj(self, dirName, jobObj):
 		# List of input files
 		input = []
 		for file in jobObj['files']:
+            inFile = file['localFile']
+            vmFile = file['destFile']
```
Z1MM32M4N: It looks like we're mixing tabs and spaces. Considering this is Python, that should probably be fixed.

ymzong: @Z1MM32M4N Good point!

Z1MM32M4N: [PEP 8][1] Says that spaces are preferred.

[1]: https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

PR #32, opened 2015-02-19
Fixed info and jobs and updated command-line client accordingly.

mihirpandya:
/pool should give a list of VMs instead of the number of VMs
admin APIs don't need courselabs


yrkumar:
this PR is unnecessary because #34 has this commit as well and it was just merged into master

PR #33, opened 2015-02-19
URI decode output files for poll

dlbucci:
Tango should now match the Autolab front-end behavior.  Untested and the REST API spec should probably be updated.


mihirpandya:
How can I test this? Should I start by setting up my staging server?

mihirpandya:
This doesn't seem to work. It still gives a 404. The encoded url looks something like:

`GET /poll/mp_test/15122-s15-pixels/mp%40out.com.out/`

dlbucci:
Is that what the client outputs?  That looks like a properly encoded URI.

mihirpandya:
This route was configured to only accept alphanumerics. Since it wasn't alphanumeric, the server returned a 404. Changed this. Will close this once we test it with the staging server.

Issue #23, opened 2015-02-03
Issue with scp failing


icanberk:
Initially mentioned here: https://github.com/autolab/Autolab/issues/169

robsimmons:
Any progress on this issue? I just did a regrade all that had a couple of failures, I've still got a bunch of misgrades for pixels I haven't fixed, and our workflow requires us to do a pretty work-intensive regrade all a week from tomorrow.

robsimmons:
I... feel like a jerk poking this issue, but the grading two assignments for 122 is kind of blocked on it.

icanberk:
Hi @robsimmons, We're really sorry for the inconvenience. Our Tango devs were a little busy with personal stuff the past week. I'll make sure that we fix this by this Friday!

Issue #31, opened 2015-02-17
/pool not returning list of free and allocated VMs


mihirpandya:

yrkumar:
#34 fixes this issue

Issue #19, opened 2015-01-26
Verbose ssh output


droh:
Instrument tango so that it can be configured to dump ssh verbose output to the logs. This would be a huge help in debugging ssh/scp issues.

yrkumar:
#34 fixes this issue by adding the -vvv flag to ssh when the config level is set to debug

PR #34, opened 2015-02-20
Fix scp

mihirpandya:
All ssh and scp actions are now retried. The number of retries and the frequency of the retries is defined in the config file. This was tested by regrading a class of 300 people.


icanberk:
This fixed #23

yrkumar:
This fixes the following issues: #19, #23, #31.

PR #39, opened 2015-03-03
Persistent job queue

icanberk:
This creates a TangoDictionary which jobQueue class uses. Depending on the configuration, TangoDictionary can be a native dictionary or a Redis based dictionary which can be shared across instances of Tango.


dlbucci:
Assignee better get to pulling.

icanberk:
Additionally, this PR refactors parts of the codebase:

1) Includes unit tests for jobQueue
2) Removes the standard modules from libs/ and creates a list of dependencies in `requirements.txt`
3) Removes object initialization from Config.py. The user only defines `VMMS_NAME` field and we initialize the right object accordingly.
4) Moves jobManager into a separate file.

icanberk:
We will try to deploy this with @yrkumar tomorrow and see what happens.

mihirpandya:
Are there any environment variables that need to be set besides  `venv`? And do I need to use supervisor to run this? I want to just run one instance of this but it has weird syntax errors that you do not seem to have, even though the code is the same.

icanberk:
You don't have to have supervisor to run this. You can just do `python tango-rest/server.py <optionally port_id>`

You also need to do `pip install -r requirements.txt` after activating venv.

icanberk:
Running Redis locally is pretty easy. After installing it, you just do `redis-server` . The default configs should be local host and default redis port.

mihirpandya:
OK so how did you test this? I'm not able to get it working with redis off, which I think should work exactly the same way as the older version.

PR #40, opened 2015-03-10
Fixes the command-line client

yrkumar:
tango-rest.py works now, with the added function "runJob" that simply uploads all the files in a given directory and adds that job to Tango. The new runJob function also takes an optional parameter "numJobs" that determines how many repeated jobs to add to Tango.

### DIFF
--- a/clients/job6/README
+++ b/clients/job6/README
@@ -1,1 +1,0 @@
-Tests resiliency to fork bombs
--- a/clients/job7/README
+++ b/clients/job7/README
@@ -1,1 +1,0 @@
-Tests resiliency to programs that exceed the VM memory size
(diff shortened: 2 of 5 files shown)


mihirpandya:
I tried the following and here's what I got:
```
./tango-rest.py -P 9091 -k <key> --prealloc --image rhel.img --num 4
>>> {"statusMsg": "Preallocate VM failed", "statusId": -1}
./tango-rest.py -P 9091 -k <key> --prealloc --image rhel.img --num 2
>>>  {"statusMsg": "VMs preallocated", "statusId": 0}
```
I think the arguments are not parsed correctly by the server, for prealloc.

PR #41, opened 2015-03-12
Tango producer consumer

mihirpandya:


dlbucci:
Maybe if we're dropping flask, we should not merge into 'newtango_withflask'?

icanberk:
I don't understand why we're switching back to Tornado now just because Flask create two processes. I'm pretty sure there is a good reason behind it. (In fact a simple googling gives this which explains why there are two processes: http://flask.pocoo.org/docs/0.10/appcontext/) It's quite clear that Tornado is not the right tool for us.

I think we should not be doing `os.fork()` from our web application in the first place. Job manager should be initiated separately. (e.g. when you want to run python celery, you need to start it from command line) Everytime we call fork ourselves we take the risk of introducing unforeseen problems + that process includes stuff from the web application which I think is a true consumer should not need.

dlbucci:
Honestly, it hasn't been made clear to me why Flask would be any better than Tornado.  If there is a problem that's because we're using a single-threaded python web framework, I don't see how switching to a different single-threaded python web framework will fix it.

icanberk:
We don't have much of a choice about threading as long as we use Python. We just don't want the single event loop model that Tornado uses because we don't need it and it slows things down. Any other framework is fine.

dlbucci:
If we're putting apache or nginx in front of it and it supports WSGI (which Tornado does), it really isn't gonna matter if it's using an event loop model or not.

icanberk:
Can we not discuss this here? You can reply to the email thread. If you feel strongly about staying with Tornado.

All I'm saying is the two process is not "erroneous" and we should not have a commit that reverts a same commit in the same branch. Purpose of this PR is the producer consumer part.

icanberk:
@mihirpandya Did we try running this yet? How did it go with a large batch?

icanberk:
So I see that the JobManager needs `vmms` and `preallocator`. Looks to me like all the fields of Preallocator, except `Preallocator.vmms`, are actually from Redis, so we can initiate a new preallocator inside JobManager and we should be fine.

I'm not sure about VMMS though since I never touched that part of the code. Is TashiSSH just a manager? Can we just initiate another TashiSSH in jobManager and use it or is it supposed to have shared fields?

icanberk:
What does `tashi.createClient` do?

icanberk:
I have an alternative suggestion here: https://github.com/autolab/Tango/pull/42

mihirpandya:
Yeah, eventually, we'll put the JobManager process as part of supervisor. It was just simpler to test it by spawning it in the beginning of the parent Tango process. And yeah, we can initiate another VMMS since it just makes a bunch of calls to control the VM with state stored in the job queue.

icanberk:
Prod/com was included as a part of #44

PR #42, opened 2015-03-12
Initial implementation of stand-alone JobManager

icanberk:
This assumes that we will start the process for JobManager separately through the command line if we have the Redis running. 

It's not tested and I'm not sure if this will work since I don't know if VMMs should be sharing any attributes. I'm pretty sure initializing Preallocators will work since `Machines` is a shared dictionary.


icanberk:
I also switched from tabs to spaces in JobManager, so it looks big but only the part after `if __name__ == "__main__":` is new.

PR #43, opened 2015-03-16
Docker vmms

mihirpandya:
Tango for OS X with docker integration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,11 +24,11 @@ COPY autodriver /usr/bin/autodriver
 RUN chmod +s /usr/bin/autodriver
 
 # Clean up
-WORK /home
+WORKDIR /home
 RUN apt-get remove -y git
 RUN apt-get -y autoremove
 RUN rm -rf Tango/
 
 # Check installation
 RUN ls -l /home
-RUN autodriver
+RUN which autodriver
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Autolab - autograding docker image
+
+FROM ubuntu:14.04
+MAINTAINER Mihir Pandya <mihir.m.pandya@gmail.com>
+
+RUN apt-get update
+RUN apt-get install -y gcc
+RUN apt-get install -y make
+RUN apt-get install -y build-essential
+
+# Install autodriver
+RUN apt-get install -y git
+WORKDIR /home
+RUN git clone https://github.com/autolab/Tango.git
+WORKDIR Tango/autodriver
+RUN ls .
+RUN make clean && make
+RUN cp autodriver /usr/bin/autodriver
+RUN which autodriver
+
+# Clean up
+RUN apt-get remove -y git
+RUN apt-get -y autoremove
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,15 +9,25 @@ RUN apt-get install -y make
 RUN apt-get install -y build-essential
 
 # Install autodriver
-RUN apt-get install -y git
 WORKDIR /home
+RUN useradd autograde
+RUN useradd autolab
+RUN mkdir autograde
+RUN chown autograde autograde
+RUN chown :autograde autograde
+RUN apt-get install -y git
 RUN git clone https://github.com/autolab/Tango.git
 WORKDIR Tango/autodriver
 RUN ls .
 RUN make clean && make
-RUN cp autodriver /usr/bin/autodriver
-RUN which autodriver
+COPY autodriver /usr/bin/autodriver
 
 # Clean up
+WORK /home
 RUN apt-get remove -y git
 RUN apt-get -y autoremove
+RUN rm -rf Tango/
+
+# Check installation
+RUN ls -l /home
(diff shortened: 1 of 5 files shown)


ymzong:
:+1:

jez:
So I think that instead of using `cat` to stick the command in the end of the Dockerfile and do other hackiness, I think this should be the general strategy:

- Create a Docker image that corresponds to all the environment setup that a given class needs.
- Copy the handin and autograde files into the Docker image somehow
  - We might be able to do this by mounting Docker Volumes
- Launch the Docker image we created by running `docker run <image name> <command to start the autodriver>`
- Write the result back to a local file (i.e., not inside the Docker container)
  - This also might be possible with Docker Volumes

jez:
That strategy uses standard tools and doesn't do weird things like piping back and forth, forcing Docker images to not terminate, ssh'ing into Docker containers, etc.

I'm not sure too much about how it would work yet, but once we get to the point where we have a "physical" VMMS, we might end up having some sort of service that sits _in front_ of Docker that communicates with Tango via SSH or HTTP or something and handles moving the files into the Volume or responding with the autograding result.

jez:
In fact, we might be able to completely eliminate the need for SSH: we could do everything with HTTP if that was easier or less prone to bugs.

mihirpandya:
In the life of a worker thread, the following happens in sequence:
* `waitVM` - check if the VM/container exists
* `copyIn` - copy the necessary files to the VM/container
* `runJob` - invoke the autodriver in the VM/container
* `copyOut` - retrieve the feedback from the autodriver

Sounds like you're implying:

* `waitVM` doesn't need to actually create a container in the context of docker
* `copyIn`, you would just create a directory on the host machine and copy the necessary files to that directory which will be used a data volume
* `runJob` would start a docker container with the necessary docker image and give it the volume to use create by `copyIn`. Something like `docker run -v <volume> <container_name>`. However, after that is done, it also needs to delete the directory/volume.
* Getting data from a docker container is actually support by docker and doesn't need any hacking. It's just `docker cp`. But we might not even need to do that since docker logs the stdout of every container to a log file stored on the host machine. So we can just use that.

This is going to create and delete a lot of directories but other than that we could actually do this. Also, creating the docker image is done once by the instructor or whoever so it doesn't really need to be a part of this whole process.

jez:
- On point three:
  - I would actually specify the volume using a `VOLUME` directive in the Dockerfile.
  - There probably would need to be some sort of cleanup of the volume, but I'm not sure whether that would mean delete the whole folder from the base (i.e., host) file system, or whether you could get by with just removing the volume or something else in Docker. I think you're on the right track though.
- On point four:
  - I'm hesitant to use logs to transfer grading data. It seems like it skirts the real issue and could even be a privacy violation if something got out of hand.
  - `docker cp` might work, or you might just be able to write a file back to the volume you mounted, and then just do a normal copy from the host file system to the host again (but outside the volume somewhere).

I don't necessarily know that it will involve creating lots of directories, but it might involve ensuring files are cleaned up.

The cool thing about using Docker for images is that we could
- Add a form in the Course Admin to let professors upload a Dockerfile
- Base our autograding Dockerfile off of theirs, something like
```Dockerfile
# Pull base image, which is configured by course instructor
FROM autolab/15122

# Add mount the directory /path/to/output/submission/locally/ as /submission/ 
# in the container, to have access to autodriver/Makefile and also capture output
VOLUME ['/path/to/output/submission/locally', '/submission']

# I'm not sure what the autodriver does, but we could do something like
ADD autodriver.sh /autodriver.sh
WORKDIR /
# Define start command
CMD ['sh', '/autodriver.sh']
# If not every course has an autodriver or if the autodriver is configurable,
# this might not work. If so, we could always push it off and have the
# instructor specify it in the autolab/15122 Dockerfile image instead
```

Don't take what I've outlined to heart too much. My experience with Docker is minimal, so I don't know best practices too much, but this is pretty reminiscent of what it looked like the last time I dabbled in it.

jez:
That last point about instructors adding their own Dockerfiles is just kind of tricky though. It would require a bit of effort because instructors would have to put extra work into creating a Dockerfile that had all the configuration they need for their course.

For people just getting started with Autolab, it'd be easier. For people used to the current VM situation, it'd be a bit jarring.

mihirpandya:
If we fix one volume then that means we cannot run multiple docker containers because then the files would clobber each other between docker containers. Having a separate directory act as a volume for each container we spin up would allow multiple jobs to run in parallel, which is nice!

Regarding your point about the logs, what issue are you talking about? And `docker cp` seems like a legitimate way of retrieving a file from the container, since it is part of the docker command-line client.

Lastly, you're right about the Dockerfile option being tricky, even though it is a great feature to allow. Alternatively, they could simply `docker push` their image to the docker registry and when we do a `docker run` with that image, it will pull the image first and then run the job on a container off of that image. So the first job will be really slow but after that it's back to normal.

icanberk:
Guys, can we continue this legit discussion on #45 . I would like to delete the `persistent_job_queue_prodcom` branch.

PR #44, opened 2015-03-18
Persistent job queue prodcom

icanberk:
The main purpose of this merge is to organize the project:
1) Gets rid of the python libraries
2) Creates a way to run the project with different port numbers
3) Includes improvements to the config file.

After these changes, Tango can be daemonized with Supervisor and run though Nginx. So on surface, nothing has changed. With Redis off, it works the same way as the current Tango.

The merge also includes two other more substantial changes:
1) Implementation of Redis based persistent memory. This feature is well tested at this point. But I'm not sure what happens when we call `tangoReset` on start twice (let's say if we started 2 tangos on the same machine)
2) Producer/Consumer pattern. Having a persistent memory let's us use this pattern. Although this seems to be working for me, it's not well tested. We should also make sure that spinning 2 Tango instances does not create 2 consumers.


icanberk:
I'm making this the master since it doesn't change anything with Redis off (tested, and deployed to greatwhite with proper nginx setup). We can continue other feature debates from new PRs to the master.

droh:
Given the persistent job queue, it's very important that there be a way to
completely reset Tango to some initial empty state, with no running VMs and
an empty job queue.

It's also not clear to me what would (should) happen to running jobs in the
Redis queue when Tango is restarted.

-- David O'Hallaron
-- Prof of CS and ECE, Carnegie Mellon University
-- GHC 7517, x8-8199
-- http://www.cs.cmu.edu/~droh

icanberk:
We have a resetTango method (https://github.com/autolab/Tango/blob/master/tangod.py#L178), but I don't think it actually resets the jobQueue. I will change it so that the jobQueue on Redis is reset as well.

Also, on the prod-com model, I think having new producers should not reset Tango. There should only be one consumer, and tango should return to initial state when there is a new consumer.

PR #45, opened 2015-03-18
Implementation of Docker vmms

icanberk:
- [x] Write documentation for how to set up Tango with Docker
  - [x] This includes writing or linking to good documentation for setting up Docker on OS X and Ubuntu
- [x] Test on...
  - [x] ... OS X (with boot2docker)
  - [x] ... Ubuntu
- [x] Allow user-specified Docker images
- [x] job1
- [x] job2
- [x] job4
- [ ] job5
- [x] job6
- [x] job7


icanberk:
`VM "boot2docker-vm" is not running.`

Is there anything like `localBootstrap.sh` ? How can I setup docker?

mihirpandya:
https://docs.docker.com/installation/mac/

vmms/dockerSSH.py
```diff
@@ -0,0 +1,323 @@
...
+        if mkdir != 0:
+            self.log.error("Failed to create directory in container %s"
+                % instanceName)
+            return -1
+
+        # Copy the input files to the input directory
+        for file in inputFiles:
+            ret = timeout(['cat', file.localFile, '|',
+                            'docker', 'exec', '-i', instanceName,
+                            'sh', '-c', 'cat > /home/autolab/' + file.destFile],
```
jez: Let's make sure we remove this part before merging.

icanberk:
I think this expects some config fields like:
Config.DOCKER_IMAGE
Config.DOCKER_IMAGE_BUILD_TIMEOUT

but they are not in config.template.py

mihirpandya:
Yeah I'm rewriting that part actually. As we discussed, there is no need to start a docker and keep it running for no reason. So only `runJob` would start a docker. And I forgot to add the docker constants in `config.template.py`; will do that too.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,323 @@
...
+            elapsed_secs = time.time() - start_time
+
+            # Give up if the elapsed time exceeds the allowable time
+            if elapsed_secs > max_secs:
+                self.log.info("Docker %s: Could not reach container \
+                    after %d seconds." % (instanceName, elapsed_secs))
+                return -1
+
+            # Give the docker container a string to echo back to us.
+            ret = dockerExec(instanceName, ['/bin/echo', echo_string])
```
icanberk: `echo_string` is not defined anywhere.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,323 @@
...
+
+            self.log.debug("Docker %s: echo returned with \
+                %d" % (instanceName, echo))
```
icanberk: You meant `%s" % (instanceName, ret))` here?

icanberk:
I keep getting
`ERROR|2015-03-19 14:10:16,727|DockerSSH|Failed to create container ilter-1004-rhel.img`

Any chance we can get more details about why it failed when logging.DEBUG?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,323 @@
+#
+# dockerSSH.py - Implements the Tango VMMS interface to run Tango jobs in 
+#                docker containers. In this context, VMs are docker containers.
+#
+import random, subprocess, re, time, logging, threading, os, sys
+import config
+from tangoObjects import TangoMachine
+
+def timeout(command, time_out=1):
```
icanberk: It looks like we use the same code in all the VMMs. Can we make this a util method?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+        instanceName = self.instanceName(vm.id, vm.name)
+        args = ['docker', 'run', '-d']
+        args.append('--name')
+        args.append(instanceName)
+        args.append(config.Config.DOCKER_IMAGE)
+        args.append('/bin/bash')
+        args.append('-c')
+        args.append('while true; do sleep 1; done')
+        ret = timeout(args, config.Config.INITIALIZEVM_TIMEOUT)
```
icanberk: So do we create the container from the Dockerfile everytime we need a new VM? I feel like it would take way too long to install all the dev binary etc and setup that Docker. Isn't it possible to create a base docker, push it to registry and build from that image in the registry?

jez: Once the image signature exists locally, it will be stored. Each step of
the dockfile build is cached, so even if you run docker build many times it
will only build it once.

We don't need our image on docker hub for this reason. We probably don't
want it there anyways, because it's not the kind of thing that is normally
made public. We might consider running our own docker registry later, so
that instructors could upload their docker files there.

Even though docker will cache build steps, it doesn't make semantic sense
to docker build the image once every time we initialize VMs. It doesn't
look like that's what you're doing here, but since Ilter brought it up I
thought I'd give you the opportunity to confirm this.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN mkdir autolab
+RUN chown autolab autolab
+RUN chown :autolab autolab
+RUN mkdir output
+RUN chown autolab output
+RUN chown :autolab output
+RUN useradd autograde
+RUN mkdir autograde
+RUN chown autograde autograde
+RUN chown :autograde autograde
```
jez: Isn't this the same as

```console
$ useradd autolab
$ useradd autograde
$ mkdir autolab output autograde
$ chown autolab:autolab autolab
$ chown autolab:autolab output
$ chown autograde:autograde autograde
```

I think this way it's a little simpler

jobManager.py
```diff
@@ -82,6 +82,9 @@ def __manage(self):
         elif Config.VMMS_NAME == "ec2SSH":
             from vmms.ec2SSH import Ec2SSH
             vmms = Ec2SSH()
+        elif Config.VMMS_NAME == "dockerSSH":
```
jez: Regretfully, this naming scheme no longer makes much sense, as we're not using SSH anymore.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+                            stderr=subprocess.STDOUT)
+    return ret
+
+def dockerExec(container, cmd, time_out=1):
+    """ docerExec - Executes `docker exec container cmd` and
+        returns output. Container is the name of the docker
+        container and cmd is a list of commands to run.
+    """
+    command = ['docker', 'exec', container, 'sh', '-c'] + cmd
+    return timeout(command, time_out)
```
jez: Is this (`dockerExec()`) an unused function?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+# User defined exceptions
+#
+
+class DockerSSH:
+    _OS_X = 'darwin'
+    LOCALHOST = '127.0.0.1'
+
+    def __init__(self):
+        """
+			Checks if the machine is ready to run docker containers.
```
jez: You have tabs (`"\t"`) here.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+        except Exception as e:
+            self.log.error(str(e))
+            exit(1)
+
+    def instanceName(self, id, name):
+        """ instanceName - Constructs a VM instance name. Always use
+        this function when you need a VM instance name. Never generate
+        instance names manually.
+        """
+        # return str(id)
```
jez: Dead code?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+    #
+    def initializeVM(self, vm):
+        """ initializeVM -  Nothing to do for initializeVM
+        """
+        return vm
+
+    def waitVM(self, vm, max_secs):
+        """ waitVM - Nothing to do for waitVM
+        """
+        return
```
jez: These two functions are why Docker is so cool XD

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def copyIn(self, vm, inputFiles):
+        """ copyIn - Create a directory to be mounted as a volume
+        for the docker containers. Copy input files to this directory.
+        """
+        instanceName = self.instanceName(vm.id, vm.name)
+
+        # Create a fresh volume
+        volume_path = config.Config.DOCKER_VOLUME_PATH + instanceName +'/'
```
jez: Did `DOCKER_VOLUME_PATH` make it into the config.template.py file?

Also, I'm not quite sure what this path is supposed to do (maybe because I can't see an example yet), but would a better name maybe be `DOCKER_VOLUME_ROOT_PATH`?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def runJob(self, vm, runTimeout, maxOutputFileSize):
+        """ runJob - Run a docker container by doing the follows:
+        - mount directory corresponding to this job to /home/autolab
+          in the container
+        - run autodriver with corresponding ulimits and timeout as
+          autolab user
+        """
+        instanceName = self.instanceName(vm.id, vm.name)
+        args = ['docker', 'run', '--name', instanceName, '-v']
```
jez: Does the `--name` flag _give_ a name to the instance being run?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+                (config.Config.DOCKER_VOLUME_PATH + instanceName, '/home/mount')]
+        args = args + [config.Config.DOCKER_IMAGE]
+        args = args + ['sh', '-c']
+
+        autodriverCmd = 'autodriver -u %d -f %d -t %d -o %d autolab &> output/feedback' % \
+                        (config.Config.VM_ULIMIT_USER_PROC, 
+                        config.Config.VM_ULIMIT_FILE_SIZE,
+                        runTimeout, config.Config.MAX_OUTPUT_FILE_SIZE)
+
+        args = args + ['cp -r mount/* autolab/; su autolab -c "%s"; \
```
jez: Rather than using `sh -c <some arbitrary command`, the preferred Docker construct for defining a default command to be run is to use `CMD`. It's basically the same thing as what you're doing, but it gets put in the Dockerfile rather than Python.

https://docs.docker.com/reference/builder/#cmd

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+                        cp output/feedback mount/feedback' % autodriverCmd]
```
jez: Rather than using `sh -c <some arbitrary command`, the preferred Docker construct for defining a default command to be run is to use `CMD`. It's basically the same thing as what you're doing, but it gets put in the Dockerfile rather than Python.

https://docs.docker.com/reference/builder/#cmd

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+
+        self.log.info('Running job: %s' % str(args))
+
+        return timeout(args, runTimeout)
```
jez: Does this work? Docker runs containers in the background when you launch them , so it returns immediately. I'm not sure if long jobs will be effectively killed.

It looks like adding a `docker timeout` command was [proposed to the Docker source](https://github.com/docker/docker/issues/1905). It's a good read.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,231 @@
...
+        self.log.info('Deleted directory %s' % volume_path)
+
+        return 0
+
+    def destroyVM(self, vm):
+        """ destroyVM - Delete the docker container.
+        """
+        instanceName = self.instanceName(vm.id, vm.name)
+        ret = timeout(['docker', 'rm', '-f', instanceName],
+            config.Config.DOCKER_RM_TIMEOUT)
```
jez: `docker rm` might block, so `timeout` should probably work here, but it'd be useful to check.

jez:
This is shaping up to be quite the feature!

Dockerfile
mihirpandya: Yup, that is better!

jobManager.py
mihirpandya: Indeed. I'm gonna go with `dockerLocal`

vmms/dockerSSH.py
mihirpandya: Yeah, it was used with the earlier model.

vmms/dockerSSH.py
mihirpandya: Something was inconsistent about the instance names so I was trying out different things. Gosh, fine, I'll get rid of it!

jez: lol so sassy :laughing:

vmms/dockerSSH.py
mihirpandya: Yeah, it gives the container a name. Then you can do `docker ps` and `docker inspect <name>` to get more information about that container. Most containers would be ephemeral but for long-running jobs this might be helpful.

vmms/dockerSSH.py
mihirpandya: Right, so I was confused about that. And the difference between `CMD` and `ENTRYPOINT` is still not very clear to me. Shed some light on that? Also, the command is different for every job because the autodriver takes different timeout, ulimit and output file size for each job.

vmms/dockerSSH.py
mihirpandya: So I'm creating directories that are to be mounted to the docker. And `DOCKER_VOLUME_PATH` is where those directories are created. The shorter the name the better, imo.

vmms/dockerSSH.py
mihirpandya: Yeah you're right. We use `timeout` to basically run a shell command in a new process with stdout pointing to `/dev/null`. But this part of the code is not to ensure that the actual job ran in `runTimeout` time. The autodriver makes sure of that. This part just ensures that the command was executed and terminated with status `0`.

jez: Oh, I should have looked at `timeout` closer.

vmms/dockerSSH.py
jez: That's my bad. I didn't realize the parameters were variable. Then what we'd want to do is put

```Dockerfile
ENTRYPOINT ["su", "autolab", "-c", "autodriver"]
```

In our Dockerfile, and then invoke docker with

```console
$ docker [options] run <docker image> [options to autodriver]
```

where `[options to autodriver]` are all the flags that you're generating up there.

If there are some useful defaults for `CMD`, then you'd want to put those in the Dockerfile as `CMD` (maybe the autodriver has defaults baked in if you specify no arguments, though, so you don't need to specify defaults).

mihirpandya: Yeah you're right. There're a bunch of steps that must happen for every container/job, which we should move to the Dockerfile. Then only the autodriver command will be passed via `docker run`.

vmms/dockerSSH.py
```diff
@@ -0,0 +1,223 @@
...
+        instanceName = self.instanceName(vm.id, vm.image)
+        volume_path = config.Config.DOCKER_VOLUME_PATH + instanceName
+        print os.listdir(volume_path)
+        print volume_path + '/feedback'
+        shutil.move(volume_path + '/feedback', destFile)
+        self.log.info('Copied feedback file to %s' % destFile)
+        
+        # Must always clean up containers in order to maintain statelessness.
+        # A solution with `docker attach` could be explored.
+        self.destroyVM(vm)
```
mihirpandya: @jez Thoughts on using `docker attach`?

mihirpandya:
@icanberk Can you pull and see if you still get the same error? FYI, before starting the Tango, you should do this from the Tango root directory:

* `docker build -t <image_name> . # don't forget the dot(.) at the end`
* Do `docker images` to see that the image you created is indeed there
* In your config, set `DOCKER_IMAGE` to `<image_name>`
* Run Tango

Gonna have to write a getting-started guide for docker...

mihirpandya:
FYI, this is how I tested with the Tango command-line client, from `Tango/clients`:

`./tango-rest.py -P 2908 -k mp_test -l docker --runJob job1/ --numJobs 20`

This command will send 20 `job1` jobs to Tango. `job1` just does an `ls -l`. You should change the port number, key and course lab to whatever you use, of course. The feedback file of a successful job looks like this:

```
Autograder [Wed Mar 25 06:55:58 2015]: Received job test_job-1:10
Autograder [Wed Mar 25 06:56:00 2015]: Success: Autodriver returned normally
Autograder [Wed Mar 25 06:56:00 2015]: Here is the output from the autograder:
---
Autodriver: Job exited with status 0
bash hello.sh
Hello from c7a6ce8c08ba:/home/autograde/autolab
I have the following files:
total 8
-rw-r--r-- 1 autograde autograde  28 Mar 25 05:55 Makefile
-rw-r--r-- 1 autograde autograde 127 Mar 25 05:55 hello.sh
```

vmms/dockerSSH.py
jez: How would you use `docker attach`? Usually you only use this if you

- want to see what your Docker container is spitting out in-progress
- interactively control the Docker image environment

I don't see how it'd be useful for `destroyVM`.

jez:
:100: :star2:

jez:
I added some TODOs to the PR description.

jez:
At some point I'll get around to revamping the Tango wiki to look more like the Autolab one, but for now just make sure that any necessary steps get documented somewhere.

mihirpandya:
It's documented in my bash history 😎

mihirpandya:
Is that a TODO list with radio buttons?? We *have* to get stuff done now, just so that I can click those dam things!

jez:
dat motivation tho

vmms/dockerSSH.py
mihirpandya: Well, I wanted to explore the possibility of reusing a container instead of destroying it every time after `copyOut`. It is not to be actually used in `destroyVM`.

jez: Hmm. While I'm not sure that we'll need to be reusing containers, we might
be able to do something similar where if we know that the next job in the
queue will use the same auto driver and makefile we copy both in and out,
but this could be a security risk.

config.template.py
```diff
@@ -60,6 +60,15 @@ class Config:
     RUNJOB_TIMEOUT = 60
     COPYOUT_TIMEOUT = 30
 
+    # Docker constants
+    BOOT2DOCKER_INIT_TIMEOUT = 5
+    BOOT2DOCKER_START_TIMEOUT = 30
+    BOOT2DOCKER_ENV_TIMEOUT = 5
+    DOCKER_IMAGE_BUILD_TIMEOUT = 300
+    DOCKER_RM_TIMEOUT = 5
+    DOCKER_IMAGE = ''
```
icanberk: Is this the path to the Dockerfile ?

Dockerfile
```diff
@@ -0,0 +1,33 @@
+# Autolab - autograding docker image
```
jez: I should have mentioned this earlier but I didn't catch it:

Normally, when a project ships a Dockerfile in the root of the repository, it signals that the app is "Dockerized," i.e., you can use Docker to actually deploy Tango. Since we're using Docker as a part of Tango (not as a way to deploy it consistently), we should move this file somewhere more appropriate (perhaps to `vmms/`?).

icanberk:
What's the docker volume pat?

vmms/dockerSSH.py
```diff
@@ -0,0 +1,221 @@
...
+                raise Exception('DOCKER_VOLUME_PATH not defined in config.')
+
+            if len(config.Config.DOCKER_IMAGE) == 0:
+                raise Exception('DOCKER_IMAGE not defined in config.')
+
+            self.log.info("Docker host IP is %s" % self.docker_host_ip)
+
+        except Exception as e:
+            self.log.error(str(e))
+            exit(1)
```
icanberk: This makes things silently die. We should get rid the line.

icanberk:
Dope. It works on my Mac :+1:

config.template.py
```diff
@@ -60,6 +60,15 @@ class Config:
...
+    DOCKER_VOLUME_PATH = ''
```
icanberk: Our general strategy with config should be things working out of the box. So I think we should ideally give a default value for these (as well as the courselabs directory) and put a setup.sh that will create the directories. If people would like to change the default values, they're free to do so.

icanberk:
What do you think about a shell script like:

```
$ boot2docker init
$ boot2docker start
$ boot2docker shellinit
$ eval "$(boot2docker shellinit)"
```

vmms/dockerSSH.py
mihirpandya: Well, it logs the error. We could also print the error to stdout. But it should die.

jobManager.py
```diff
@@ -82,6 +82,9 @@ def __manage(self):
...
+            from vmms.dockerSSH import DockerSSH
+            vmms = DockerSSH()
```
icanberk: If system needs to exit, I think there should be one Rescue here that catches all the initialization errors, not just docker, and exit.

icanberk: Basically something was wrong in my config and it was exiting, but it was hard to pin down where it was going wrong. Printing something to stdout will also be very helpful with that.

mihirpandya:
Yeah, we can either have a shell script or make it part of the constructor, under line 74. But I thought since the user will have to follow instructors to download docker and boot2docker, they might as well follow the instructions there to get docker running.

config.template.py
mihirpandya: You're right. I added some default constants.

icanberk: :yen:

jobManager.py
mihirpandya: Hmm, ok. Yeah exiting on a higher level makes more sense. But we have to make sure that it exits because of a problem that affects all of Tango and not just because of one job. It shouldn't be too big of a problem because worker threads have their own exception handler and so any stray job won't bring the whole system down.

config.template.py
mihirpandya: Did you just make it rain?

icanberk: Yeah, but in Yen

icanberk: :cn:

mihirpandya: 👲💃

jez:
Yeah, I'd rather leave Docker configuration steps to Wiki documentation, rather than have scripts that do them. It forces them to understand what's going on, which makes it easier in case they have a difference in the way their system is set up.

jez:
As far as I can tell, it's only documentation and tests now. Good work @mihirpandya :star:

config.template.py
```diff
@@ -60,6 +60,16 @@ class Config:
...
     COPYOUT_TIMEOUT = 30
 
+    # Docker constants
+    BOOT2DOCKER_INIT_TIMEOUT = 5
+    BOOT2DOCKER_START_TIMEOUT = 30
+    BOOT2DOCKER_ENV_TIMEOUT = 5
+    DOCKER_IMAGE_BUILD_TIMEOUT = 300
+    DOCKER_RM_TIMEOUT = 5
+    # Must be absolute path with trailing slash
+    # Default value of '*'' points this path to /path/to/Tango/volumes/
```
icanberk: this currently points out to " /path/to/Tango/volumes//" (two (( at the end) which causes an error

icanberk:
Werks on Ubuntu.

PR #48, opened 2015-03-19
Don't re-upload pre-existing files

yrkumar:
Stops upload of files that already exist on the server so that there is no race condition when we scp those files to VMs.


mihirpandya:
Thanks for creating a separate object for file exists in status :D

PR #64, opened 2015-04-07
Adding getImages in VMMS

mihirpandya:
The VMMS library needs another function: `getImages`. This will return a list of images that the VMMS is capable of booting its VM/container with. We should remove tashi-specific code from Tango in `validateJob`.


mihirpandya:
@yrkumar 

Tested with:
`./tango-rest.py -P 2908 -k mp_test --prealloc --image rhel.img --num 3`
`./tango-rest.py -P 2908 -k mp_test --pool --image rhel`
`./tango-rest.py -P 2908 -k mp_test -l jobs --runJob job2/ --image rhel.img`

Note that pool takes in the pool name (without the .img) but addJob and prealloc take in the image name (with the .img). This wouldn't matter once we use docker.

PR #67, opened 2015-04-13
Added pseudo job-ordering by ID back into Tango. Fixes #55 for tashiSSH

### LINKED ISSUE #55: Tango jobs are not dispatched in any sort of pseudo ordering
Jobs should roughly be dispatched in the order which they came. Exceptions are only when some jobs are not schedulable.

yrkumar:
Tested with tashiSSH. needs testing with Redis and Docker @mihirpandya.

### DIFF
--- a/config.template.py
+++ b/config.template.py
@@ -23,7 +23,7 @@ class Config:
     PORT = 3000
 
     # Log file. Setting this to None sends the server output to stdout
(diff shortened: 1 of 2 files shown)


mihirpandya:
Looks good, except for all the Travis issues. @ymzong plz halp

PR #69, opened 2015-04-14
Distributed docker

mihirpandya:


vmms/distDocker.py
```diff
@@ -0,0 +1,313 @@
...
+# host machines must be able to run docker and be accessible
+# by SSH. The IP address of the host machine is stored in the
+# `domain_name` attribtue of TangoMachine.
+#
+
+import random, subprocess, re, time, logging, threading, os, sys, shutil
+import config
+from tangoObjects import TangoMachine
+
+def timeout(command, time_out=1):
```
jez: So I didn't catch this for the last PR, but is there a reason why the timeout function is not in a library somewhere? It looks like `timeout` and `timeoutWithReturnStatus` are just copied between all the vmms's.

mihirpandya: Yeah there's no reason. Just hadn't gotten down to separating the two.

vmms/distDocker.py
```diff
@@ -0,0 +1,315 @@
...
+              "-o", "StrictHostKeyChecking=no",
+              "-o", "GSSAPIAuthentication=no"]
+
+    def __init__(self):
+        """ Checks if the machine is ready to run docker containers.
+        Initialize boot2docker if running on OS X.
+        """
+        try:
+            self.log = logging.getLogger("DistDocker")
+            self.hosts = ['54.186.238.205', '54.68.89.235']
```
jez: I'm sure you know this, but this will want to be factored out into a config parameter eventually.

mihirpandya: Yeah. I wanted to think a little more about that. It would be cool to add and remove host machines live instead of having to restart Tango every time. And to do that, maybe we store the path to a separate `hosts_config` file in the main config file and install a signal handler in `DistDocker` to reload that config file. But then there're two config files, which is messy and confusing...

icanberk:
Btw, I think we should switch to the same version control mechanism on Tango. (ie. Having the `develop` branch)

vmms/distDocker.py
icanberk: Yeah I also commented about this previously. We can have a vmmsUtils or something like that.

vmms/distDocker.py
icanberk: Maybe have a default list in the config file and let the users add new ones through the cmd line client? (Like I think we do for VMs right now)

mihirpandya:
Is there any way to disable Travis since it doesn't work Docker? All the red crosses don't mean anything...

ymzong:
I think there's no way to disable an already existing PR, but since I already disabled it on Travis, new commits/PRs will not go through Travis.

mihirpandya:
Looks like new commits are still going through. Don't know about PRs.

mihirpandya:
Oh and @icanberk I agree with the develop branch idea. How can we do it?

ymzong:
I would say the first step is to establish a stable version. Once you are there, simply branch out `develop`.

PR #75, opened 2015-07-01
Make the JobManager thread a daemon

cg2v:
The JobManager threas was not set up as a daemon, so it was not possible
to shut down JobManager or non-redis tango-rest with SIGINT or SIGTERM


cg2v:
Not quite right - the standalone JobManager now exits immediately

cg2v:
And it's back to not exiting on Ctrl-C. maybe jobManager.py's main should just call signal.pause()

A more invasive possibility is to only use a thread when the JobManager is runnung inside TangoREST

mihirpandya:
The way we normally terminate the job manager is to send it a SIGTSTP (ctrl-z) signal first and then to kill it:

* Ctrl-Z
* `kill %1`

yrkumar:
Closed by b654c6e3fba59182b2a4cd7b4cb543a5bc727f0a

PR #76, opened 2015-07-09
Don't hardcode container interface in Docker vmms

cg2v:
Instead of putting all the file copying and su-ing in the docker run
command constructed by the vmm, include an interface script in the
container and use it as the container entrypoint. The only thing
passed from the vmm to the container are the resource limit parameters.

The script interface might be changed to pass in the volume name and/or the name of the output file so the script doesn't even to know that.

In addition, rewrite the dockerfile so it (1) includes fewer RUN lines, and
(2) does not need to fetch Tango from github in order to build autodriver;
Instead, the autodriver source is copied to the container using the ADD
instruction


mihirpandya:
I agree this is a better approach than putting all the constants in the vmms itself. So let's iterate on this idea.

The only problem I noticed so far is that on jobs that should timeout, `runJob` exits with `EXIT_OSERROR` instead of `EXIT_TIMEOUT`. This is a problem because an OS error represents a VM that should no longer be used. So that triggers a catastrophic retry, meaning that the VM is destroyed, recreated, added back to the pool and the job is retried. The reason we are getting this error is because the autodriver is not able to kill the job and perform its clean up. I've pasted a sample output file.

Also, why did you move from Python to Perl? Seems like you had some Python 2 vs 3 compatibility issue? I think most of the developers are more familiar with Python and would like to keep things in Python as much as possible unless there is a very good reason to switch.

```
Autograder [Sun Aug  2 07:09:12 2015]: Received job test_job-1:2
Autograder [Sun Aug  2 07:10:35 2015]: Internal error: Unable to complete job after 2 tries. Pleae resubmit
Autograder [Sun Aug  2 07:10:35 2015]: Job status: waitVM=None copyIn=0 runJob=3 copyOut=0
Autograder [Sun Aug  2 07:10:35 2015]: Here is the output from the autograder:
---
Autodriver: Job timed out after 60 seconds
bash hello.sh
Hello from 7e59f0478238:/home/autograde/autolab
Autodriver: Gave up killing user processes at line 383
```

cg2v:
The python->perl change was a result of what was in the sample container, assuming that the containers used in production would be similarly minimal, and trying to pick something that would work on both ubuntu and rhel based containers. As the sample Dockerfile is not actully representative, I would certainly be inclined to go back to the python script, as it's easier to read.

cg2v:
The timeout mismatch happened because there were zombie processes after autodriver did its first kill, and the wrapper, which is pid 1, didn't wait for them. Apparently, when a shell is pid 1, it does end up calling wait() and preventing such zombies from accumulating. I added a background thread to the wrapper which calls wait until autodriver exits (or wait fails). 
i also decided to use exec directly, so the parent process gets direct access to the exit status.

cg2v:
the distDocker change has not been tested

PR #77, opened 2015-07-15
Try to return VM to queue if worker gets an error

cg2v:
if worker encounters an exception, replace the VM before
exiting.


cg2v:
Part of the ba9ecec merge

PR #79, opened 2015-07-15
RFC: Use less subprocess call

cg2v:
I found some suspect uses of subprocess.call in the source that could potentially be replaced with os or shutil functions. Is there a reason for the use of subprocess.call other than to suppress exceptions if the operations fail?


cg2v:
Rebased. Still only a suggestion for discussion.

yrkumar:
Closed by ce97dbea71c3913274ac50b34db25a0dab398e52 and 2b0bdeee28f64a6cd4523ede2e0e67f9fba15be3

PR #86, opened 2015-09-11
Check filename and md5 hash on file uploads

yrkumar:
Solves "caching" issue by only rejecting file uploads if the same file exists on the server **with the same filename**. Note that this also changes the REST API spec for /open from returning a list of dictionaries to returning just a simple dictionary. This will require a change on frontend that @ymzong is implementing.

### DIFF
--- a/restful-tango/tangoREST.py
+++ b/restful-tango/tangoREST.py
@@ -146,12 +146,12 @@ class TangoREST:
         """ computeMD5 - Computes the MD5 hash of given files in the
         given directory
         """
-        result = []
+        result = {}
         for elem in os.listdir(directory):
             try:
                 body = open("%s/%s" % (directory, elem)).read()
                 md5hash = hashlib.md5(body).hexdigest()
-                result.append({'md5': md5hash, 'localFile': elem})
+                result[elem] = md5hash
             except IOError:
                 continue
         return result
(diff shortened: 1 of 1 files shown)


restful-tango/tangoREST.py
```diff
@@ -301,12 +302,11 @@ def upload(self, key, courselab, file, body):
             labPath = self.getDirPath(key, courselab)
             try:
                 if os.path.exists(labPath):
+                    fileMD5 = hashlib.md5(body).hexdigest()
+                    filesInDir = self.computeMD5(labPath)
+                    if file in filesInDir and filesInDir[file] == fileMD5:
+                        return self.status.file_exists
```
icanberk: So even if one file/directory is the same we return self.status.file_exists and stop the whole upload process?

yrkumar: No, so previously we were not uploading the file if the md5 hash of the contents matched the md5 hash of one of the files already in the directory. This means that a file with a different name but the same contents would not be uploaded. 

Now, we only reject the upload if **both the filename and md5 hash** are the same. We actually check both the filename and the hash on the frontend, and we originally thought that was enough but it wasn't so we want to do this check in the backend as well.

yrkumar: Also, this is only for one call to upload (i.e. one file upload), it doesn't stop the whole upload process for a job at all because the client can call upload many times.

icanberk:
Looks good to me. You can merge.

yrkumar:
closed by 050e5fcd3a6e5af1644cf606c9a2eb9bb0fa5b42

PR #89, opened 2015-09-14
Remove localSSH VMMS. Fixes #88.

### LINKED ISSUE #88: LocalSSH VMMS should be removed
The `localSSH` VMMS is basically deprecated in favor of `localDocker`, so we should remove it

yrkumar:
`localSSH` is deprecated in favor of `localDocker` because of the comparatively difficult setup process for `localSSH` as well as the restriction that it only works on linux.


icanberk:
LGTM.

PR #95, opened 2015-09-17
Remove unnecessary files and misc clean up

yrkumar:
Pretty self-explanatory. Tango needs a lot of cleanup work - legacy scripts and dependencies still remain in the codebase from a long time ago. This is only the first cleanup PR that I'm working on.


yrkumar:
Added 2nd commit that moves some logic from the API level to the server level where it belongs. I'd suggest looking at the changes from the first commit and then the changes from the second commit independently

clients/tango-rest.py
```diff
@@ -4,13 +4,10 @@
 # tango-rest.py - Command line client for the RESTful Tango.
 #
 
-
 import os
 import sys
 
 sys.path.append('/usr/lib/python2.7/site-packages/')
-sys.path.append('gen-py')
-sys.path.append('/usr/share/Tango-prod/lib/requests-2.2.1/')
```
icanberk: This is so ghetto

icanberk:
LGTM!

yrkumar:
Closed by 55fe85f2b1899db7efe320754a90b818c88f9289

PR #96, opened 2015-09-17
Change array of key-values to python dictionary

### LINKED ISSUE #81: json output is hard to parse
The JSON output of getInfo and getPool uses JSON Arrays of ["a=X", "b=Y"] instead of Objects {"a":X, "b": Y}. This makes them non trivial to parse, since the client has to dive into the string rather then letting a JSON library do all the work for it. 

Is there a reason for doing things this way?

nayak16:
Resolves #81

### DIFF
--- a/tangod.py
+++ b/tangod.py
@@ -182,16 +182,6 @@ class TangoServer:
         stats['copyout_errors'] = Config.copyout_errors
         stats['num_threads'] = threading.activeCount()
         
-        #stats = []  
-        #stats.append("elapsed_secs=%d" % (time.time() - Config.start_time))
-        #stats.append("job_requests=%d" % Config.job_requests)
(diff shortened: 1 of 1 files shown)


yrkumar:
LGTM, but this may require a frontend change as well to make sure we don't break Autolab integration. @ymzong @jez @icanberk?

ymzong:
@ykumar I'm on it

ymzong:
@yrkumar  I'm on it

yrkumar:
We can merge this, but wait until @ymzong's change goes into Autolab to pull to production.

ymzong:
@yrkumar @nayak16 Front-end fix is ready!

nayak16:
Merged f779953

PR #98, opened 2015-09-20
Set start_time to system time when tango is started

### LINKED ISSUE #70: `elapsed_secs` field in `info` endpoint is wrong
When I hit `/info` endpoint in the API, the `elapsed_secs` variable is shown as the current epoch time (e.g. `1429372581`) instead of the actual number of seconds elapsed.

May you forgot to save the time when Tango was started?

nayak16:
Resolves #70

### DIFF
--- a/config.template.py
+++ b/config.template.py
@@ -3,6 +3,7 @@
 #
 
 import logging
+import time
 
 # Config - defines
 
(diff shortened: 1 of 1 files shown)


icanberk:
LGTM.

nayak16:
Merged f736923

PR #99, opened 2015-09-22
Add start script that kills previous instance of server and starts new instance

nayak16:
I know this script was deleted initially. However I found it easier, if you are changing server specific code, to use this script to restart the server and start another fresh instance in the background. Speeds up development. Let me know what you guys think.


yrkumar:
2 things:
- the `source scripts/sourcelib` line is no longer necessary
- will that `grep` kill the first result that contains `server` in the list of processes, even if it isn't Tango?

nayak16:
I'll remove that line. And yes actually it will. I'll make the regex more specific

yrkumar:
@nayak16 should I close this PR or do you still want to merge this in?

nayak16:
Close this PR. Script isn't necessary.

PR #102, opened 2015-10-10
Remove duplicated work in jobManager

yrkumar:
This work in `jobManager` was already being done in `TangoServer`


nayak16:
LGTM :+1:

PR #103, opened 2015-10-27
Ensure job manager pulls jobs out of Redis queue in order of ids

nayak16:


nayak16:
@yrkumar once you look at the issue we discussed offline, we can move forward with this.

yrkumar:
@nayak16, the jobManager now assigns jobs on startup if the Redis job queue is already populated

PR #105, opened 2015-11-05
Add job to deadJobs before adding to liveJobs

### LINKED ISSUE #104: Crash Vulnerability in Job Manager
Currently, jobs are removed from the liveJobs queue to the deadJobs queue. If the JobManager goes down before a job is added to the deadJob queue, the job will be lost. 
Possible solution: only remove from liveJobs queue when adding to deadJobs queue succeeds with no error.

nayak16:
Fixes #104

### DIFF
--- a/jobQueue.py
+++ b/jobQueue.py
@@ -262,11 +262,11 @@ class JobQueue:
         if str(id) in self.liveJobs.keys():
             self.log.info("makeDead| Job is in the queue")
             status = 0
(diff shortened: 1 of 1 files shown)


nayak16:
Is this good to merge? @yrkumar

nayak16:
Merged 56ba5bbc78323199b2d7f1cdf7d21b4d0ca6d75b

PR #106, opened 2015-11-19
Refresh the job dict in Redis after editing the job retries

yrkumar:
VM becomes unrecoverable after OS error and the job churns through VMs, rescheduling the job infinitely. The issue is that after retrying the job, we update `job.retries` but don't refresh the job entry in Redis, so the next time the job is assigned, `job.retries` remains the old version.

### DIFF
--- a/jobQueue.py
+++ b/jobQueue.py
@@ -241,14 +241,14 @@ class JobQueue:
         self.queueLock.acquire()
         self.log.debug("unassignJob| Acquired lock to job queue.")
         job = self.liveJobs.get(jobId)
-        self.log.info("unassignJob|Unassigning job %s" % str(job.id))
-        job.makeUnassigned()
         if job.retries is None:
             job.retries = 0
         else:
             job.retries += 1
             Config.job_retries += 1
 
+        self.log.info("unassignJob|Unassigning job %s" % str(job.id))
+        job.makeUnassigned()
         self.queueLock.release()
         self.log.debug("unassignJob| Released lock to job queue.")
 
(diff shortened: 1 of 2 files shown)


nayak16:
Why was this not an issue in the native job queue? Was it because you are still modifying the same object in the native job queue, whereas in Redis, you have to serialize and add it back in?

yrkumar:
I believe so - I think this is required because the Redis queue is no longer in memory and so the objects that are referenced are different when we get an object from Redis whereas they would be the same when getting an object from an in-memory queue.

nayak16:
Makes sense :+1:

droh:
Yash,
Just to double-check: Did you eliminate the retries altogether if the
autodriver returns an OS error? If Tango receives this error, it should
kill the VM, start a new replacement VM, and then terminate the job, with
no retries.


-- David O'Hallaron
-- Prof of CS and ECE, Carnegie Mellon University
-- GHC 7517, x8-8199
-- http://www.cs.cmu.edu/~droh

yrkumar:
nope, I didn't do that. this code currently will retry two times before failing. let me update the PR.

droh:
Can you do that now? For the storage class, the retries on OS error are a
big problem because each retry takes about 3 minutes, which has the
potential to clog up the queue.


-- David O'Hallaron
-- Prof of CS and ECE, Carnegie Mellon University
-- GHC 7517, x8-8199
-- http://www.cs.cmu.edu/~droh

yrkumar:
ok, two changes with the new update:
1. synced Redis after updating retries so that `job.retries` gets reflected in the remote object
2. edited condition to stop retrying when encountering return code of 3, and instead simply replace the VM and make the job dead since it could have damaged the VM

PR #109, opened 2015-11-26
Return the correct result on create

cg2v:
when open creates a directory, return an empty hash/object,
not an empty array, so as to behave the same as when the directory
already exists (as changed in #86)


yrkumar:
Closed by 63e3b438e81552fe6f990e30852ba06f127eb0da

PR #111, opened 2015-12-03
Reorganization

yrkumar:
This PR has 3 major changes:
1. Adds module structure so we can use import statements in subdirectories without having to programmatically add to the system `PATH`. This means that we now start Tango using the following command from the shell: `python -m restful-tango.server`. These changes are in 268075e. 
2. Removes duplicate dependencies in Tango (i.e. `jobManager` requires only `jobQueue` to initialize since `preallocator` and `vmms` are stored in `jobQueue`). These changes are in 85b435c.
3. Renames `tangod.py` to `tango.py` and adds extra logging information that will help identify a job by its ID in the logs. 

Tested locally with `localDocker` vmms to make sure everything is still working.

### DIFF
--- a/jobQueue.py
+++ b/jobQueue.py
@@ -7,11 +7,11 @@
 # JobManager: Class that creates a thread object that looks for new
 # work on the job queue and assigns it to workers.
 #
-import time, threading, logging
+import threading, logging, time
 
 from datetime import datetime
-from config import Config
 from tangoObjects import TangoDictionary, TangoJob
+from config import Config
 
 #
 # JobQueue - This class defines the job queue and the functions for
@@ -79,7 +79,7 @@ class JobQueue:
             self.log.info("add|JobQueue is full")
             return -1
         self.log.debug("add|Gotten next ID: " + str(job.id))
-        self.log.info("add|Unassigning job %s" % str(job.id))
+        self.log.info("add|Unassigning job ID: %d" % (job.id))
         job.makeUnassigned()
         job.retries = 0
 
@@ -100,9 +100,8 @@ class JobQueue:
         self.queueLock.release()
         self.log.debug("add|Releasing lock to job queue.")
 
-        self.log.info("Added job %s:%d to queue" % (job.name, job.id))
-        self.log.info("Job id: " + str(job.id))
-        self.log.info("Job details: " + str(job.__dict__))
+        self.log.info("Added job %s:%d to queue, details = %s" % 
+            (job.name, job.id, str(job.__dict__)))
 
         return str(job.id)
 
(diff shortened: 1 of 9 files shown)


jobQueue.py
```diff
@@ -79,7 +79,7 @@ def add(self, job):
             self.log.info("add|JobQueue is full")
             return -1
         self.log.debug("add|Gotten next ID: " + str(job.id))
-        self.log.info("add|Unassigning job %s" % str(job.id))
+        self.log.info("add|Unassigning job ID: %d" % (job.id))
```
nayak16: Minor: not sure if this is the better, the job is being unassigned not the job ID. Maybe "Unassigning job #%d" would be better?

tango.py
```diff
@@ -71,30 +67,29 @@ def __init__(self):
             from vmms.distDocker import DistDocker
             vmms = DistDocker()
 
-        self.vmms = {Config.VMMS_NAME: vmms}
-        self.preallocator = Preallocator(self.vmms)
+        self.preallocator = Preallocator({Config.VMMS_NAME: vmms})
```
nayak16: Why can't TangoServer just have an attribute `vmms` like the following: `self.vmms = self.preallocator.vmms`?

jobQueue.py
yrkumar: True, I just wanted the job ID in the log message so that when we are viewing the logs, we can associate this specific message with a job ID. As it was previously, we would not be able to tell which job this message was associated with.

nayak16: Didn't the previous statement still show the job ID? `self.log.info("add|Unassigning job %s" % str(job.id))`

nayak16:
Also, why is `tangod.py` changed to `tango.py`? Does `tangod.py` imply it is a daemon?

jobQueue.py
yrkumar: you're right. what I meant to say above was that the log messages elsewhere tend to have this established format of using `job ID: #` and this change was intended to formalize that.

yrkumar:
yeah, `tangod` implies it is a daemon, which it isn't. In fact, it really isn't even a server since we added the restful layer on top of it.

tango.py
yrkumar: we can do that, but it seems unnecessary because we can access it via preallocator as `self.preallocator.vmms`.

yrkumar:
@nayak16 I'm just aiming to simplify things here... the more the co-dependencies, the more complex the code gets and that's something we'd like to avoid.

nayak16:
Gotcha, just trying to see the motivation behind it

yrkumar:
@nayak16 any issues with merging this?

nayak16:
Looks good to merge :+1:

PR #113, opened 2015-12-07
Smart hashing of files on receiving open/upload request

yrkumar:
Two changes
1. Stop returning file hashes on `open`
2. Check file hashes only if a file with the same name is found on `upload`


icanberk:
This looks good to me :100:

PR #114, opened 2015-12-19
Dockerize Tango

nayak16:
Needs to be tested on multiple machines (i.e. OSX) to ensure everything works correctly. 
See documentation for instructions: https://github.com/autolab/Tango/wiki/Suggested-Deployment


yrkumar:
@nayak16, I haven't tried this out yet - will do that soon, but in the meantime can you add a `README.md` to the `deployment/` directory and use that to explain the things you have in the `.txt` files here? This will organize the directory better and also will force a better explanation of the various text files.

nayak16:
Yeah I can do that along with a couple other changes I realized I need to make. Let me know whenever you test it, should still work with the current version.

yrkumar:
@nayak16 I tested this and it looks ready to go.

Issue #121, opened 2016-02-26
Improve suggested deployment of Tango


yrkumar:
This issue consists of many parts:
- [x] Add nginx and docker to the list of programs managed by supervisor
- [x] Make supervisor start these programs automatically when docker container is run
- [x] Expose ports on docker container so that Tango service is accessible from the outside 
- [ ] Use host's docker installation instead of docker-in-docker

fanpu:
Resolved with docker-compose installation

PR #123, opened 2016-02-26
Add nginx and docker to supervisor with instructions to autostart

yrkumar:
This addresses some parts of #121. More specifically, it adds nginx and docker to supervisor. Exposing the ports of the Tango service in the docker container simply involves adding extra params when running the docker container so that has been added to the [suggested deployment documentation](https://github.com/autolab/Tango/wiki/Suggested-Deployment).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,12 +11,12 @@ WORKDIR /opt
 
 # Move all code into Tango directory
 ADD . TangoService/Tango/
-RUN mkdir /volumes
+WORKDIR /opt/TangoService/Tango
+RUN mkdir volumes
+
+WORKDIR /opt
 
 # Install dependancies
-# supervisor installed and works
-
-
 RUN apt-get update && apt-get install -y \
 	nginx \
 	curl \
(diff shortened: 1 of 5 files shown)


yrkumar:
@nayak16 does this look good to merge?

Dockerfile
```diff
@@ -53,22 +52,21 @@ RUN chmod +x /usr/local/bin/wrapdocker
 
 # Define additional metadata for our image.
 VOLUME /var/lib/docker
-CMD ["wrapdocker"]
```
nayak16: Why was this removed? @yrkumar

yrkumar: `docker ps` seems to work inside the docker container even without running this wrapper script so I feel more comfortable leaving it out. It seems to be a wrapper to start docker, which won't work in the `Dockerfile` anyways since you can only run one command at the end of the `Dockerfile` to be run in the container. 

As an alternative, I can set `wrapdocker` as the process to start in `supervisor` instead of Docker itself, which will ensure that `wrapdocker` runs (and thus will start Docker). Is this preferred?

nayak16:
Let's merge this. It looks good

yrkumar:
Closed by b452d05848e09628720e57a8232d53a1d2eadb25

PR #124, opened 2016-03-17
Ec2 bug fixes

yunfanye:
I tried the setup guide and testing guide, but it only works after fixing the following four bugs:
1. The indent for destroyVM
2. Add a temporary getImages method (it actually doesn't need this method)
3. Change t1.micro to t2.micro. It seems that there isn't t1.micro instance.
4. When we ssh to an AWS, the user name can be ec2-user or root instead of ubuntu. But this is hard-coded. See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html.

### DIFF
--- a/vmms/ec2SSH.py
+++ b/vmms/ec2SSH.py
@@ -227,35 +227,35 @@ class Ec2SSH:
                 if (elapsed_secs > max_secs):
                     return -1
 
-            # The ping worked, so now wait for SSH to work before
-            # declaring that the VM is ready
-            self.log.debug("VM %s: ping completed" % (vm.name))
-            while(True):
+        # The ping worked, so now wait for SSH to work before
+        # declaring that the VM is ready
+        self.log.debug("VM %s: ping completed" % (vm.name))
+        while(True):
 
-                elapsed_secs = time.time() - start_time
+            elapsed_secs = time.time() - start_time
 
-                # Give up if the elapsed time exceeds the allowable time
-                if elapsed_secs > max_secs:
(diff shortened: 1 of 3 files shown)


vmms/ec2SSH.py
```diff
@@ -380,3 +380,9 @@ def existsVM(self, vm):
             if inst.instances[0].id is vm.ec2_id:
                 return True
         return False
+
+    def getImages(self):
+        """ getImages - Lists all images in TASHI_IMAGE_PATH that have the
+        .img extension
+        """
+        return ["test.img"]
```
nayak16: Maybe just return an empty list? Is there a particular reason a test.img is returned?

yunfanye: Sorry about this hacking, but empty list wouldn't work. The list returned by getImages is used for validation, i.e. whether the image specified by the user is in the list. Also, Tango will do some other works regarding this name.

After a second thought, I believe that test.img is badly named. Perhaps default.img is better, which means using the default AMI specified in the config.py file. (Perhaps the corresponding wiki pages https://github.com/autolab/Tango/wiki/Testing-Tango and https://github.com/autolab/Tango/wiki/Set-up-Amazon-EC2-VMMS should change as well.) 

FYI, an alternative is to use a bypass list and change all references to the function getImages. But that means for every future service provider, we should update that list.

nayak16:
Merged d71ea5382d1062cd5eb82a2553a7a255a984e0fe

PR #125, opened 2016-04-11
Allow Autolab front-end to specify the AWS credentials that are used in auto-grading

yunfanye:
Changes proposed in this PR:
-  Allow Autolab front-end to specify the AWS credentials that are used in auto-grading
-  Back-compatibility preserved (new features are effective only when accessKeyId and acceessKey is passed in)
-  change restful clients accordingly

### DIFF
--- a/clients/tango-rest.py
+++ b/clients/tango-rest.py
@@ -201,7 +201,6 @@ def tango_addJob():
         requestObj['accessKeyId'] = args.accessKeyId
         requestObj['accessKey'] = args.accessKey
 
-        print "Sent request to %s:%d/addJob/%s/%s/ \t jobObj=%s" % (args.server, args.port, args.key, args.courselab, json.dumps(requestObj))
         response = requests.post(
             'http://%s:%d/addJob/%s/%s/' %
             (args.server,
@@ -209,6 +208,7 @@ def tango_addJob():
              args.key,
              args.courselab),
             data=json.dumps(requestObj))
+        print "Sent request to %s:%d/addJob/%s/%s/ \t jobObj=%s" % (args.server, args.port, args.key, args.courselab, json.dumps(requestObj))
         print response.content
 
     except Exception as err:
--- a/clients/tango-rest.py
+++ b/clients/tango-rest.py
@@ -83,6 +83,15 @@ parser.add_argument(
     '--notifyURL',
     help='Complete URL for Tango to give callback to once job is complete.')
 
+# add for aws student accounts
+parser.add_argument(
+    '--accessKeyId', default='',
+    help='AWS account access key ID')
+
+parser.add_argument(
+    '--accessKey', default='',
+    help='AWS account access key content')
+
 
 def checkKey():
     if (args.key is None):
@@ -185,9 +194,14 @@ def tango_addJob():
         requestObj['max_kb'] = args.maxsize
         requestObj['output_file'] = args.outputFile
         requestObj['jobName'] = args.jobname
+        
         if (args.notifyURL):
             requestObj['notifyURL'] = args.notifyURL
+        
+        requestObj['accessKeyId'] = args.accessKeyId
+        requestObj['accessKey'] = args.accessKey
 
+        print "Sent request to %s:%d/addJob/%s/%s/ \t jobObj=%s" % (args.server, args.port, args.key, args.courselab, json.dumps(requestObj))
         response = requests.post(
             '%s:%d/addJob/%s/%s/' %
             (args.server,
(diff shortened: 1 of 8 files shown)


clients/tango-rest.py
```diff
@@ -185,17 +194,21 @@ def tango_addJob():
...
         requestObj['output_file'] = args.outputFile
         requestObj['jobName'] = args.jobname
+        
         if (args.notifyURL):
             requestObj['notifyURL'] = args.notifyURL
+        
+        requestObj['accessKeyId'] = args.accessKeyId
+        requestObj['accessKey'] = args.accessKey
 
+        print "Sent request to %s:%d/addJob/%s/%s/ \t jobObj=%s" % (args.server, args.port, args.key, args.courselab, json.dumps(requestObj))
```
nayak16: Any reason you placed this before the `requests.post()`?

jobManager.py
```diff
@@ -44,32 +44,48 @@ def run(self):
...
+
                 try:
                     # Mark the job assigned
                     self.jobQueue.assignJob(job.id)
-
-                    # Try to find a vm on the free list and allocate it to
-                    # the worker if successful.
-                    if Config.REUSE_VMS:
-                        preVM = vm
+                    # if the has specified an account
```
nayak16: Nitpick: typo in comment

vmms/ec2SSH.py
```diff
@@ -102,6 +110,13 @@ def instanceName(self, id, name):
         """
         return "%s-%d-%s" % (config.Config.PREFIX, id, name)
 
+    def keyPairName(self, id, name):
```
nayak16: Any reason this function is `instanceName()`. They seem to have a similar function

clients/tango-rest.py
yunfanye: The intuition is that we still want to know what the request is, if it happens to fail and gets caught by the try statement.

vmms/ec2SSH.py
```diff
@@ -244,7 +299,7 @@ def waitVM(self, vm, max_secs):
             # If the call to ssh returns timeout (-1) or ssh error
             # (255), then success. Otherwise, keep trying until we run
             # out of time.
-            ret = timeout(["ssh"] + Ec2SSH._SSH_FLAGS +
+            ret = timeout(["ssh"] + self.ssh_flags +
```
nayak16: Thanks for this change! It's better style to have a `ssh_flags` property rather than "hardcode" the `Ec2SSH_SSH_FLAGS` property everywhere

worker.py
```diff
@@ -47,7 +47,10 @@ def detachVM(self, return_vm=False, replace_vm=False):
         or not in the pool (replace_vm). The worker must always call
         this function before returning.
         """
-        if return_vm:
+        # job-owned instance, simply destroy after job is completed
+        if self.job.accessKeyId:
```
nayak16: This if branch action is the same as the else statement. They both do `safeDestroyVM`. Any reason for this?

vmms/ec2SSH.py
yunfanye: Yes. You are right. They are the same. Frankly speaking, I just copied the `instanceName()` function. The only difference is that `instanceName()` is used to name the instance, but `keyPairName()` is used to name the keypair associated with the instance. If you believe that this makes the code redundant, I can delete it.

clients/tango-rest.py
nayak16: If the request fails then the log message may be something like this:
`Sent request to...`
`Failed to send request to...`
Both log messages will show the same details about what the log message is. Thus, even if the request fails we'll know what request it tried to make. The "Sent" and "Failed" is also slightly confusing because the "Sent" implies it succeeded.

yunfanye: Oh...yes. Change it back.

nayak16:
@yrkumar Can you look at this asap? I want this merged before the end of the semester

jobManager.py
```diff
@@ -44,32 +44,48 @@ def run(self):
 
     def __manage(self):
         self.running = True
+        # job-associated instance id
+        nextId = 10000
```
yrkumar: where does this 10000 number come from? should this be `Config.MAX_JOBID` instead?

jobManager.py
```diff
@@ -44,32 +44,48 @@ def run(self):
...
+                    # if the job has specified an account
+                    # create an VM on the account and run on that instance
+                    if job.accessKeyId:
+                        from vmms.ec2SSH import Ec2SSH
+                        vmms = Ec2SSH(job.accessKeyId, job.accessKey)
+                        newVM = copy.deepcopy(job.vm)
+                        newVM.id = nextId
+                        preVM = vmms.initializeVM(newVM)
+                        if nextId > 99999:
+                            nextId = 1000
```
yrkumar: these hardcoded numbers here should probably be changed to some constants that are initialized in `config.py`

yunfanye: No. It is not job id but an intern id, a unique identifier, for VMs. It could also be generate via random(), but I use this in an attempt to be consistent with the original code, which uses hardcoded number between 1000 and 9999. To be more specific, I make a small change.

nayak16:
I'm going to merge this in today @yunfanye Sorry about the delay

PR #129, opened 2017-01-25
Improved stability of distDocker and added RAM limit option to tango

daviddob:
Changes proposed in this PR:
-  Add RAM limit to tango containers
      -This adds protection in the event a student runs code that infinitely mallocs memory and hoses the      physical server
     -Configurable in the config.py
-  Improved the stability of distDocker and tested with 5 server cluster (using REUSE_VMs=True)


vmms/distDocker.py
```diff
@@ -96,6 +92,18 @@ def __init__(self):
             self.log.error(str(e))
             exit(1)
 
+    def readHosts(self):
+        f = open(self.HOSTS_FILE, 'r')
```
nayak16: Why have it in a hosts file instead of `config.py`?

vmms/distDocker.py
```diff
@@ -68,12 +67,8 @@ class DistDocker:
 
     _SSH_FLAGS = ["-q", "-o", "BatchMode=yes" ]
     _SSH_AUTH_FLAGS = [ "-i", os.path.join(os.path.dirname(__file__), "id_rsa"),
-              "-o", "StrictHostKeyChecking=no",
-              "-o", "GSSAPIAuthentication=no"]
-    _SSH_MASTER_FLAGS = ["-o", "ControlMaster=yes",
```
nayak16: Why are these removed? Don't we want to use multiplexing? Maybe a better flag may be `ControlMaster=auto` which will use the master connection if one exists, and create one otherwise.

vmms/distDocker.py
```diff
@@ -172,16 +175,9 @@ def copyIn(self, vm, inputFiles):
         instanceName = self.instanceName(vm.id, vm.image)
         volumePath = self.getVolumePath(instanceName)
 
-        if vm.use_ssh_master:
```
nayak16: Same as above. Why aren't we using an ssh master?

hosts
```diff
@@ -1,2 +1 @@
-54.186.238.205
-54.68.89.235
\ No newline at end of file
+jarvis.cse.buffalo.edu
```
nayak16: This shouldn't be checked into git. It would be better to define hosts in `config.py` as mentioned below.

daviddob:
Updated according to suggestions.

PR #130, opened 2017-01-27
Added ram limit to docker containers and patched distDocker vmms

daviddob:
Changes proposed in this PR:

Add RAM limit to tango containers
-This adds protection in the event a student runs code that infinitely mallocs memory and hoses the physical server
-Only applied to localdocker and distdocker vmms

-Configurable in the config.py
Improved the stability of distDocker and tested with 5 server cluster (using REUSE_VMs=True)


vmms/distDocker.py
```diff
@@ -117,6 +118,15 @@ def initializeVM(self, vm):
         """ initializeVM -  Assign a host machine for this container to 
         run on.
         """
+        self.hostLock.acquire()
+        host = self.hosts[self.hostIdx]
+        self.hostIdx = self.hostIdx + 1
+        if self.hostIdx >= len(self.hosts):
```
nayak16: Can just do `self.hostIdx = (self.hostIdx + 1) % len(self.hosts)`

nayak16:
Don't think you meant to push the last commit here?

daviddob:
Pardon? Which commit?

Thanks,
Dave

nayak16:
This one: [added unzip to autograding image](https://github.com/autolab/Tango/pull/130/commits/31c53076e20dafa1b43364c9da903e9dc6e69fa5)

daviddob:
Hmm didn't realize it would modify the PR. Updating the dockerfile is just
to adapt it to meet our requirements for the jobs it will be running

Dave

nayak16:
Yeah makes sense. You probably pushed to the same branch, hence it updated here. However, we shouldn't be committing specific language installations into the main autograding image Dockerfile. It's fine if you guys use it in your deployment, however.

PR #145, opened 2017-11-19
Return MD5 of existing files on open

### LINKED ISSUE #137: REST Tango /open don't return Files list
### Expected Behavior
According to documentation, when I am getting into /open I suppose to see:
{ 
  "statusMsg": <string>, 
  "statusId": <int>,
  "files": { <fileName1> : <md5hash1>, <fileName2> : <md5hash2> ... },
}
### Actual Behavior
When getting into /open request, it returns only StatusMsg and StatusId, Files are always empty even if there is any file in courselab directory:

{ 
  "statusMsg": "Found directory", 
  "statusId": "0"
  "files": { },
}
### Steps to Reproduce the Behavior
Just add few lines of code in Open method and add CoumputeMD5 method:
DIR: tango/restful-tango/tangoREST.py
```python
    def computeMD5(self, directory):
        """ computeMD5 - Computes the MD5 hash of given files in the
        given directory
        """
        result = {}
        for elem in os.listdir(directory):
            try:
                body = open("%s/%s" % (directory, elem)).read()
                md5hash = hashlib.md5(body).hexdigest()
                result[elem] = md5hash
            except IOError:
                continue
        return result

      def open(self, key, courselab):
        """ open - Return a dict of md5 hashes for each input file in the
        key-courselab directory and make one if the directory doesn't exist
        """
        self.log.debug("Received open request(%s, %s)" % (key, courselab))
        if self.validateKey(key):
            labPath = self.getDirPath(key, courselab)
            try:
                if os.path.exists(labPath):
                    self.log.info(
                        "Found directory for (%s, %s)" % (key, courselab))
                    statusObj = self.status.found_dir
                    statusObj['files'] = self.computeMD5(labPath)
                    return statusObj
                else:
                    outputPath = self.getOutPath(key, courselab)
                    os.makedirs(outputPath)
                    self.log.info(
                        "Created directory for (%s, %s)" % (key, courselab))
                    statusObj = self.status.made_dir
                    statusObj["files"] = {}
                    return statusObj
            except Exception as e:
                self.log.error("open request failed: %s" % str(e))
                return self.status.create(-1, str(e))
        else:
            self.log.info("Key not recognized: %s" % key)
            return self.status.wrong_key
```

zyx-billy:
Fixes #137 (involves reverting #113).

#113 changed the behavior of `open` so that it always returns an empty md5 dict. This means the client will always send all their files each time. Interestingly, md5 checking is done during upload requests. Not only is this inefficient, it is not even needed: If the client already sent us their latest version of the file, there's no point in checking md5 at this point, just save it. The entire point of md5 was so that the client wouldn't need to send us everything.

Changes proposed in this PR:
- Send back hashes of existing files during an `open` request. This way the client can compare checksums and only send over new and updated files.
 - The status `file_exists` is removed since it's not useful or even worth it to compare checksums during an upload request. The client should be able to make the choice of what files it want uploaded. And it makes no difference in the end if the client uploaded an existing file. (In fact, if by some odd chance, two files resulted in the same md5, this would be the only way it could get updated).

### DIFF
--- a/restful-tango/tangoREST.py
+++ b/restful-tango/tangoREST.py
@@ -28,7 +28,6 @@ class Status:
         self.found_dir = self.create(0, "Found directory")
         self.made_dir = self.create(0, "Created directory")
         self.file_uploaded = self.create(0, "Uploaded file")
-        self.file_exists = self.create(0, "File exists")
         self.job_added = self.create(0, "Job added")
         self.obtained_info = self.create(0, "Found info successfully")
         self.obtained_jobs = self.create(0, "Found list of jobs")
@@ -102,18 +101,19 @@ class TangoREST:
         labPath = self.getDirPath(key, courselab)
         return "%s/%s" % (labPath, self.OUTPUT_FOLDER)
 
-    def checkFileExists(self, directory, filename, fileMD5):
-        """ checkFileExists - Checks if a file exists in a
-            directory
+    def computeMD5(self, directory):
+        """ computeMD5 - Computes the MD5 hash of given files in the
+        given directory
         """
+        result = {}
         for elem in os.listdir(directory):
-            if elem == filename:
-                try:
-                    body = open("%s/%s" % (directory, elem)).read()
-                    md5hash = hashlib.md5(body).hexdigest()
-                    return md5hash == fileMD5
-                except IOError:
-                    continue
+            try:
+                body = open("%s/%s" % (directory, elem)).read()
+                md5hash = hashlib.md5(body).hexdigest()
+                result[elem] = md5hash
+            except IOError:
+                continue
+        return result
 
     def createTangoMachine(self, image, vmms=Config.VMMS_NAME,
                            vmObj={'cores': 1, 'memory': 512}):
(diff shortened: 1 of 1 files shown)


cg2v:
There is some value in not repeatedly overwriting the common files on each grade request. If you do overwrite them, you risk having the vmms copying a partial file if another tango thread is writing at the same time the vmms is reading (or running scp). You could fix that by using a .NEW file and renaming if you really wanted to.

The file list was removed from tango because it caused performance problems when assignment directories had alot of items in them. That's where the excessive computations server side occur.

zyx-billy:
I see. In that case, I agree we should not haphazardly overwrite existing files. We should probably implement a solution for replacing files since it does occur even right now, admittedly very rarely. However, at the same time, this seems like something we could minimize by letting the client know what files we already have so they don't even need to send it. If the only problem with this is too much computation during `open`, have we considered storing the calculated md5s (perhaps in each courselab directory?), and just send that over during `open` (and obviously, they'll be updated each time `upload` gets something new).

PR #149, opened 2018-04-17
tangoREST.py: Fix notifyURL typo.

lonerz:


zyx-billy:
Hi, were you trying to use tango-cli? It turns out that we can't just change the name here since other services currently depend on this name (e.g. In Autolab, we use this key when adding jobs). Instead of changing the interface, it may be a better idea to just fix the cli.

lonerz:
Closing this as stale

PR #156, opened 2019-01-22
Fix regression during file upload

loicgelle:
Regression caused by commit 05e4d0f375afccd35b85ebd5ef8b7f96000f6324. Files are
copied from a temp file to the final destination as the file has not been fully
uploaded and flushed. That yielded to grading corrupted or empty files.

Fixes #154


cg2v:
this change is correct.  i've been doing too much ruby lately

PR #157, opened 2019-02-01
Clojure setup

rzolau:
Added setup for Clojure to the (default?) Dockerfile.
Fixed some minor typos in comments.


fanpu:
Closed due to staleness and because the default Dockerfile is too bloated

Issue #153, opened 2018-07-13
Autograding. Error message: 795: unexpected token at '---'


ronzalo:
### Context
AutoGrading a quiz with a ruby script and localDocker

### Expected Behavior
Autodriver ouput writed correctly

### Actual Behavior
Ruby `puts` output isn' writted

### Possible solution 
In https://github.com/autolab/Tango/blob/master/vmms/localDocker.py#L141, use `autolab > output/feedback 2>&1` instead `autolab &> output/feedback`

s-wallace:
I can reproduce this problem as well; the localDocker call seems to work fine with an autograding image built on ubuntu 14.04, but not on 18.04 here are a few more details:

 - sh is invoked by localDocker which is dash 0.5.7 on ubuntu 14.04 and dash 0.5.8 on ubuntu 18.04
 - `autolab &> output/feedback` isn't strictly supported in dash (see https://wiki.ubuntu.com/DashAsBinSh) 

to reproduce:

1. build a docker autograding image based on ubuntu 14.04 (the current Dockerfile in Tango/vmms will work)
2. build a second based on unbuntu 18.04
3. run the command issued by localDocker, mine looks like:

```
docker run --rm --name grader -v <pathwithsubmission>:/home/mount autograding_image sh -c \
'cp -r mount/* autolab/; su autolab -c "autodriver -u 100 -f 1048576 -t 10 -o 524288 \
autolab &> output/feedback"; cp output/feedback mount/feedback'
```

This command works fine for me on a 14.04 image, and produces no output on a 18.04 image

4.  run the proposed command:
```
docker run --rm --name grader -v <pathwithsubmission>:/home/mount autograding_image sh -c \
'cp -r mount/* autolab/; su autolab -c "autodriver -u 100 -f 1048576 -t 10 -o 524288 \
autolab > output/feedback 2>&1"; cp output/feedback mount/feedback'
```

This produces correct output for both images.

fanpu:
I am running into an issue where if this fix is used, autodriver seems to interpret `2>&1` as a token to parse instead:

```autolab@aede15bcd16e:/home$ autodriver -u 100 -f 104857600 -t 180 -o 1024000 autolab 2>&1 output/feedback
autodriver: Too many arguments```

This results in autograding not being performed at all.

Looking to see if there is an alternative solution

fanpu:
Confirming that this works

fanpu:
Fix now in master

Issue #138, opened 2017-09-25
NoneType Exception When Queuing Jobs


chriswailes:
### Expected Behavior

When using Redis with VM ReUse enabled the server should continue to process jobs.

### Actual Behavior

After some combination of job submissions takes place the jobManager will produce the following error:

```
DEBUG|2017-09-24 16:58:32,713|JobQueue|get| Acquired lock to job queue.
DEBUG|2017-09-24 16:58:32,724|JobQueue|get| Released lock to job queue.
Starting the stand-alone Tango JobManager
Traceback (most recent call last):
  File "jobManager.py", line 125, in <module>
    jobs.run()
  File "jobManager.py", line 45, in run
    self.__manage()
  File "jobManager.py", line 105, in __manage
    self.jobQueue.makeDead(job.id, str(err))
AttributeError: 'NoneType' object has no attribute 'id'
```

### Steps to Reproduce the Behavior

Run the Tango job manager with Redis and VM ReUse.

### Additional Details

The error appears to be the result of `getNextPendingJobReuse` returning `(None, None)` at jobManager.py:66.  This then results in the next line trying to retrieve a job using `None` as an ID.  This results in an exception and the variable `job` holding the value `None`.  Next, the exception handler at jobManager.py:105 attempts to mark the job as dead.  Unfortunately, because `job` is `None` the above error is raised.

chriswailes:
It turns out the "solution" to this issue is a necessary but entirely undocumented command.  Tango keeps track of "pools" (that aren't actually pools) of VMs and attempts to "allocate" machines from there.  For some reason the pool for the first Docker image name given is automatically created for you, but you can't use a different Docker image without manually creating them using the following command:

```
python ./clients/tango-cli -k <key> -l <lab name> --prealloc --image <image name> --num <pool size>
```

With VM re-use disabled (having it enabled was causing problems for me) the pool size is a useless parameter.  Tango will simply spawn as many jobs as it can as fast as it can.

victorhuangwq:
Indeed there seems to be an issue with VM Reuse, and it still has not be resolved. Bumping this thread to have devs to notice

victorhuangwq:
Hi, a quick update for those who are viewing this thread later, read the TLDR below if you don't want to read through the explaination. 

Firstly looking at this line in the codebase, the docker containers are already destroyed after copying out the feedback
https://github.com/autolab/Tango/blob/master/vmms/localDocker.py#L159

[...]

PR #160, opened 2019-12-27
Python 3 Upgrade

fanpu:
### Plan:
Unfortunately, we have almost no test coverage so regressions must be detected by hand. Our goal is to port Tango to Python 3 while still maintaining Python 2 compatibility.

1. Follow Futurize porting guide (http://python-future.org/automatic_conversion.html) with manual inspection. For each stage, create PRs that merge into this PR so it is clearer which changes are being introduced at which stage for accountability. 
2. Test Tango CLI on Python 3 manually (maybe write some tests?)
3. Test Tango on Python 3 with Autolab
4. Test Tango CLI on Python 2 manually
5. Test Tango on Python 2 with Autolab

Tested with all CLI commands given in https://autolab.github.io/docs/tango-cli/

This is now ready for review!


fanpu:
I have reviewed the stage 1 fixes and they LGTM

jobManager.py
```diff
@@ -62,9 +67,16 @@ def __manage(self):
...
                 job = self.jobQueue.get(id)
+
+                # job could no longer exist if it was completed by someone else
+                if job == None:
+                    continue
+
                 if not job.accessKey and Config.REUSE_VMS:
                     id, vm = self.jobQueue.getNextPendingJobReuse(id)
                     job = self.jobQueue.get(id)
+                    if job == None:
```
victorhuangwq, line 78: this fix does seem to address the issue mentioned here https://github.com/autolab/Tango/issues/138.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Start with empty ubuntu machine
-FROM ubuntu:15.04
+FROM ubuntu:xenial
```
victorhuangwq, line 2: there will be issues when obtaining output from autograder if changed to 18.04 based on https://github.com/autolab/Tango/issues/153

there are relevant patch in there to address it. Has that been added?

jobQueue.py
```diff
@@ -134,7 +137,7 @@ def remove(self, id):
         self.log.debug("remove|Acquiring lock to job queue.")
         self.queueLock.acquire()
         self.log.debug("remove|Acquired lock to job queue.")
-        if str(id) in self.liveJobs.keys():
+        if str(id) in list(self.liveJobs.keys()):
```
victorhuangwq, line 140: Is self.liveJobs a dictionary?

In that case it would be more suitable to change it to `if str(id) in self.liveJobs` (w/o calling keys) for python 3
instead of converting it into a list first.

https://stackoverflow.com/questions/1602934/check-if-a-given-key-already-exists-in-a-dictionary

preallocator.py
```diff
@@ -29,7 +31,7 @@ def __init__(self, vmms):
     def poolSize(self, vmName):
         """ poolSize - returns the size of the vmName pool, for external callers
         """
-        if vmName not in self.machines.keys():
+        if vmName not in list(self.machines.keys()):
```
victorhuangwq, line 34: same comment with regards to the key situation in jobqueue

tangoObjects.py
```diff
@@ -317,23 +323,23 @@ def set(self, id, obj):
         self.dict[str(id)] = obj
 
     def get(self, id):
-        if str(id) in self.dict.keys():
+        if str(id) in list(self.dict.keys()):
             return self.dict[str(id)]
```
victorhuangwq, line 327: same comment as above

vmms/distDocker.py
```diff
@@ -8,6 +8,8 @@
 # `domain_name` attribtue of TangoMachine.
```
victorhuangwq, line 8: As you have changed the default docker image to xenial the issue here would happen https://github.com/autolab/Tango/issues/138, 

do update with the patch from here
https://github.com/mtoupsUNO/Tango/commit/b97bc1e1eaa000644d4e030e44e1c96341c2d120

jobQueue.py
fanpu: It's either a TangoNativeDictionary or TangoRemoteDictionary (which exports the same interface) which is not iterable so the `x in y` syntax would not work here

preallocator.py
fanpu: self.machines here is also either a TangoNativeDictionary or TangoRemoteDictionary

tangoObjects.py
fanpu: Works here, this is using python dict

vmms/distDocker.py
fanpu: Patch applied

Dockerfile
fanpu: xenial is actually 16.04, but I just changed it to 18.04. This Dockerfile is for the Tango Dockerfile, I did not modify the autograding Dockerfile in this PR which is still on 14.04 (should fix that soon though)

fanpu:
@victorhuangwq Thanks for your comments! I've rectified the issues you raised, do look at it again when you're available :D

tangoObjects.py
```diff
@@ -317,23 +323,23 @@ def set(self, id, obj):
...
     def get(self, id):
-        if str(id) in self.dict.keys():
+        if str(id) in self.dict:
             return self.dict[str(id)]
         else:
             return None
 
     def keys(self):
-        return self.dict.keys()
+        return list(self.dict.keys())
```
victorhuangwq, line 332: doesn't it return a list here? 

Wouldn't calling list again on keys() be redundant in the other lines I have highlighted above? Such as in
`if vmName not in list(self.machines.keys()):`

tangoObjects.py
```diff
@@ -279,15 +284,16 @@ def set(self, id, obj):
...
             unpickled_obj = self.r.hget(self.hash_name, str(id))
             obj = pickle.loads(unpickled_obj)
             return obj
         else:
             return None
 
     def keys(self):
-        return self.r.hkeys(self.hash_name)
+        keys = map(lambda key : key.decode(), self.r.hkeys(self.hash_name))
+        return list(keys)
```
victorhuangwq, line 296: It seems to be returning a list when keys are called on the TangoRemoteDictionary.

preallocator.py
victorhuangwq: refer to comment below

cg2v:
This should not block the upgrade, since I believe the current code to be correct, but most if not all of the keys() calls in the code do not require the list form of the keys (you only need to listify the keys if you either want to use [] to access elements of the key list, or if you will alter the dict while iterating over the keys). Furthermore, a more readable way to do
`if vmName in self.machines.keys():`

would be to implement __contains__ in the Tango*Dictionary classes themselves and then call

`if vmName in self.machines:`

instead. TangoRemoteDictionary could use HEXISTS to implement __contains__ and not have to fetch everything.

jobManager.py
```diff
@@ -117,7 +129,7 @@ def __manage(self):
         tango = TangoServer()
         tango.log.debug("Resetting Tango VMs")
         tango.resetTango(tango.preallocator.vmms)
-        for key in tango.preallocator.machines.keys():
+        for key in list(tango.preallocator.machines.keys()):
```
cg2v, line 132: Another unnecessary list()

preallocator.py
```diff
@@ -44,7 +46,7 @@ def update(self, vm, num):
         of machines as necessary.
         """
         self.lock.acquire()
-        if vm.name not in self.machines.keys():
+        if vm.name not in list(self.machines.keys()):
```
cg2v, line 49: another unnecessary list()

fanpu:
> This should not block the upgrade, since I believe the current code to be correct, but most if not all of the keys() calls in the code do not require the list form of the keys (you only need to listify the keys if you either want to use [] to access elements of the key list, or if you will alter the dict while iterating over the keys). Furthermore, a more readable way to do
> `if vmName in self.machines.keys():`
> 
> would be to implement `__contains__` in the Tango*Dictionary classes themselves and then call
> 
> `if vmName in self.machines:`
> 
> instead. TangoRemoteDictionary could use HEXISTS to implement `__contains__` and not have to fetch everything.

@cg2v  This is a great suggestion, I will add it as a future enhancement instead of adding that to this PR to avoid making this PR bigger than it already is.

Thank you and really appreciate the spotting of other redundant wrappers of `list` as well!

victorhuangwq (approved):
**Testing**: Tested Tango3 branch by grading various assignments through Autolab, and it works as expected.

I have verified that:
1. Patch from https://github.com/autolab/Tango/issues/153 has been applied
2. Unnecessary lists, as mentioned by both @cg2v  and I, has been removed
3. Dockerfile is now back on 18.04 (change removed)

This pull request LGTM :)

victorhuangwq:
@fanpu Before you merge in however, please create an alternate Python 2 branch, preferably called `master-python-2` to hold the old code. 

At the same time, here are the list of things we should also do alongside this PR. I can help with it as well. 

- [ ] Update the description on the readme under the section **Python 3 Upgrade.** I do believe the wording that was used in the Autolab readme regarding rails 4 support is good. 
- [ ] Add a new release, marking this as a new version of Tango. (v2.0.0 maybe?)

PR #161, opened 2019-12-27
Python 3 Upgrade Stage 2 Changes

fanpu:
Current status:
We have only done the stage 1 fixes in python3-upgrade, so Tango does not work with Python 3 yet. 

Outstanding work:
- python package version issues (pip install doesn't work right now bc of conflicts) - need to find correct versions of packages that support python3
- most automatic fixers in stage2 not applied yet
- unsafe fixes definitely not applied yet

Porting strategy:
- See http://python-future.org/automatic_conversion.html on general porting guideline
- I will advise that we apply one fix at a time (i.e futurize -f lib2to3.fixes.fix_basestring -w *.py */*.py) and manually inspecting the git diff, so as not to be overwhelmed by all the fixes at once

### DIFF
--- a/config.template.py
+++ b/config.template.py
@@ -2,12 +2,13 @@
 # config.py - Global configuration constants and runtime info
 #
 
+from builtins import object
 import logging, time
 
 # Config - defines
 
 
-class Config:
+class Config(object):
     #####
     # Part 1: Tango constants for developers
     #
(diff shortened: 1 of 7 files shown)


fanpu:
Closing this as I would rather make individual PR for major fixes

PR #167, opened 2020-05-15
Porting the autograder image to alpine linux

wlnirvana:
When getting suggested buffer size with`sysconf(_SC_GETPW_R_SIZE_MAX)` on some OSes (e.g. Alpine linux with musl,  FreeBSD, etc.), -1 is returned, causing autodriver to exit. This PR sets the buffer size to 1024 based on the [POSIX doc][1] 

[1]:https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/getpwnam.html#top

### DIFF
--- a/autodriver/Makefile
+++ b/autodriver/Makefile
@@ -6,7 +6,7 @@ OBJS = autodriver.o
 all: autodriver
 
 autodriver: $(OBJS)
-	$(CC) $(LDFLAGS) -o autodriver $(OBJS)
+	$(CC) -o autodriver $(OBJS) $(LDFLAGS)
 	sudo chown root autodriver
 	sudo chmod +s autodriver
 
--- a/autodriver/Makefile
+++ b/autodriver/Makefile
@@ -7,8 +7,6 @@ all: autodriver
 
 autodriver: $(OBJS)
 	$(CC) -o autodriver $(OBJS) $(LDFLAGS)
-	sudo chown root autodriver
-	sudo chmod +s autodriver
 
 autodriver.o: autodriver.c
 
--- a/autodriver/autodriver.c
+++ b/autodriver/autodriver.c
@@ -130,9 +130,16 @@ static int parse_user(char *name, struct passwd *user_info, char **buf) {
     long bufsize;
     int s;
 
-    if ((bufsize = sysconf(_SC_GETPW_R_SIZE_MAX)) < 0) {
-        ERROR_ERRNO("Unable to get buffer size");
-        exit(EXIT_OSERROR);
+    errno = 0;
+    bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+    if (bufsize < 0) {
+        // POSIX doc: need to set bufsize if -1 returned and errno unchanged
+        if (bufsize == -1 && errno == 0) {
+            bufsize = 1024;
+        } else {
+            ERROR_ERRNO("Unable to get buffer size");
+            exit(EXIT_OSERROR);
+        }
     }
 
     if ((*buf = malloc(bufsize)) == NULL) {
--- a/autodriver/autodriver.c
+++ b/autodriver/autodriver.c
@@ -398,7 +398,7 @@ static void cleanup(void) {
     // We are currently in ~user.
     // (Note by @mpandya: the find binary is in /bin in RHEL but in /usr/bin
     // in Ubuntu)
-    char *find_args[] = {"find", "/usr/bin/find", ".", "/tmp", "/var/tmp", "-user", 
+    char *find_args[] = {"env", "find", ".", "/tmp", "/var/tmp", "-user",
         args.user_info.pw_name, "-delete", NULL};
     if (call_program("/usr/bin/env", find_args) != 0) {
         ERROR("Error deleting user's files");
--- a/vmms/Dockerfile_alpine
+++ b/vmms/Dockerfile_alpine
@@ -1,0 +1,19 @@
+FROM alpine:3.11
+
+WORKDIR /home
+RUN apk add bash make coreutils procps findutils gcc libc-dev argp-standalone && \
+    mkdir -p autodriver && \
+    cd autodriver && \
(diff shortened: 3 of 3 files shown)


wlnirvana:
Note that the modifications made so far are not enough for basing the autograding docker image on alpine linux, primarily because busybox does not satisfy [the setuid requirement][1] of the `autodriver` binary.

`autodriver` is designed to [run as user `autolab`][3] instead of `root`, but relies on [setuid being set][4] for itself to gain root privileges. This enables the binary to use utilities like [`mv`][2], [`chown`][5], etc. to manipulate data not belonging to user `autolab`.

Unfortunately, alpine's core utility busybox will [automatically drop privileges][6] for [`mv`][7], causing [this line][2] to fail. I tried to bypass the issue with a manually compiled qucik-and-dirty replacement for `mv`. It worked, but `autodriver` failed at `chown` then.

I haven't checked all the commands used by `autodriver`, but I doubt they will work. Meanwhile, manually replacing all the binaries just seems too much effort and error-prone.

[1]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/autodriver/autodriver.c#L261
[2]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/autodriver/autodriver.c#L292
[3]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/vmms/localDocker.py#L146
[4]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/vmms/Dockerfile#L24
[5]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/autodriver/autodriver.c#L307
[6]:https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/Config.in#L198
[7]:https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/coreutils/mv.c#L25

fanpu:
Thanks for spotting this issue, and the incredibly detailed comments! The changes for autodriver.c looks good to me, and right now the main issue is with the setuid for the autodriver when you are running on Alpine. I understand this PR is still a WIP, so I will ignore the changes to Makefile first. I will be following along, since getting autograding woring on Alpine is a huge performance boost and something that I haven't been able to get successfully working in the past before (due to the fact that Autolab was still on Rails 4 back then and there were dependencies that were too old and didn't exist in libraries available for Alpine) and so it's something that I'm definitely very excited about if it's actually possible.

wlnirvana:
I've made a prototype base image that works with the hello demo lab in my local environment. Changes committed mainly are:

1. Reordered the autodriver compiler flags, in particular `LDFLAGS`. This flag is empty on Ubuntu but that's fine because argp is built-in on Ubuntu. However, alpine needs a separate `argp-standalone` package to be installed and explicitly linked, so the linking flag order matters. To maintain compatibility with Ubuntu, the flag is only set in the [dockerfile][1].
2. Removed `sudo` from the makefile, because  alpine does not have `sudo` installed by default. And more importantly, it is not the `Makefile`'s responsibility (except in full-fledged `make install` command) to manage the permission settings, but can be done in dockerfile as with the case of [current ubuntu docker file][2].
3. Minor updates in `autodriver.c` to port it to alpine linux.
4. A sample alpine based dockerfile with gcc and libc-dev undeleted, so that it can run against the hello lab.

Also note that instead of cloning the entire Tango repo, I used `wget` to download the updated autodriver source code **FROM MY PERSONAL FORK** of tango.

[1]:https://github.com/autolab/Tango/pull/167/files#diff-9f9cbd4d444c85bf29308f072bf79f20R10
[2]:https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/vmms/Dockerfile#L24

fanpu:
Our CMU sysadmin in charge of managing our deployments mentioned that a way to get around the Busybox dropping privileges issue is to set the uid to 0 inside autodriver.c, either at the start of the program, or alternatively inside call_program after fork is performed and before execv. As of now the removal of the setuid bit on autodriver inside the Makefile is the most contentious part of this PR as I'm not sure if it may break any existing forms of deployment, so if the workaround can be adopted that would be much safer

wlnirvana:
Manually setting uid to 0 worked in my previous experiments (a PoC one, not directly on autodriver.c), but I was concerned that code changes in autodriver.c would break the system or expose security vulnerabilities (note that this program will run `make` and user submitted code) on other testing backends like EC2, ssh, etc. That's why I opted for a more conservative approach.

I don't really get the Makefile point, though. For dockerised grading images,`autodriver` binary is copied over like in the following dockerfile, where permissions **must be set again** on the copy destination rather than source:

https://github.com/autolab/Tango/blob/e56a6c6497d4ab71e62f6c6dad347fa404dc2476/vmms/Dockerfile#L23-L24

This effectively invalidates any permission manipulation in makefile.

I'm not sure about other deployments like EC2, local machines, etc. But as long as the `autodriver` executable is copied, it should not break the system.

fanpu:
I agree with you that the setuid in the Makefile is pointless for Docker deployments. Apparently the setuid thing was something that was implemented a long time ago and might have been to have something to do with deployments on (Tashi VMs)[https://www.cs.cmu.edu/~jclopez/ref/tashi-acdc09.pdf], because they might have thought that SSH-ing into Tashi VMs and executing autodriver as a non-root user with setuid is safer than allowing root login, but to be honest no one on the current team really knows why it is there. I know this isn't entirely helpful in resolving the issue but I hope it might shed some light on the situation.

wlnirvana:
That makes sense. Any update on the setuid issue?

wlnirvana:
Closing this PR as I am synchronizing my master branch with Tango upstream master. Changes committed are checked into [feature/porting-autograding-to-alpine](https://github.com/wlnirvana/Tango/tree/feature/porting-autograding-to-alpine) branch.

PR #168, opened 2020-06-02
Dockerfile OS update

fanpu:
Changes Ubuntu version to 18.04 as 15.04 is long out of EOL.

The environment variable is added because tzdata has an interactive prompt for ease of usage by default but this stalls the Docker installation. The usage of the env var ensures that the interactive prompt is disabled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Start with empty ubuntu machine
-FROM ubuntu:15.04
+FROM ubuntu:18.04
 
 MAINTAINER Autolab Development Team "autolab-dev@andrew.cmu.edu"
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Start with empty ubuntu machine
-FROM ubuntu:15.04
+FROM ubuntu:18.04
```
oliverli, line 2: Could this work with Ubuntu 20.04 LTS? I think using 20.04 now will save us a fair bit of trouble down the road when 18.04 reaches EOL

fanpu: I've run into issues before where some packages I need were not actually available on Ubuntu 20.04 yet as it's relatively new (released April this year). 18.04 will be supported until 2023 - I think we can consider upgrading a little further down the road once the 20.04 ecosystem stabilises

PR #169, opened 2020-06-13
Make volumes paths explicit to avoid assuming cwd

fanpu:
In the Tango deployment instructions for Docker on Ubuntu 18.04, os.cwd() is unable to find the current directory of the Tango installation. I'm guessing most likely that the problem arose because of a difference in how supervisord or the OS determines the current working directory of a service. Previously, the cwd was probably inherited from where the supervisor daemon was started, but this implicit behavior is probably undefined behavior and we shouldn't be relying on it.

This PR forces the volume directory to be explicitly specified.

### DIFF
--- a/config.template.py
+++ b/config.template.py
@@ -40,6 +40,12 @@ class Config:
     # "tashiSSH", and "ec2SSH"
     VMMS_NAME = "localDocker"
 
+    # Update this to the 'volumes' directory of your Tango installation if
+    # Docker is being used.
+    # It must be an absolute path with trailing slash, i.e
+    # /opt/TangoService/Tango/volumes/
+    DOCKER_VOLUME_PATH = '/opt/TangoService/Tango/volumes'
+
     #####
     # Part 2: Constants that shouldn't need to change very often.
     #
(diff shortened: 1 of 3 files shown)


fanpu:
> LGTM. Removes mentioned undefined behavior. However, I was wondering if the need to specify volume directory was mentioned anywhere / added in the tango documentation (under the Docker section).

I'll be addressing this shortly. Thanks for reviewing this!

config.template.py
```diff
@@ -40,6 +40,12 @@ class Config:
     # "tashiSSH", and "ec2SSH"
     VMMS_NAME = "localDocker"
 
+    # Update this to the 'volumes' directory of your Tango installation if
+    # Docker is being used as the VMMs.
+    # It must be an absolute path with trailing slash, i.e
+    # /opt/TangoService/Tango/volumes/
+    DOCKER_VOLUME_PATH = ""
```
wlnirvana, line 47: I understand this has been merged into master already, but why deleting the default volume path instead of appending the trailing slash to the previous string `/opt/TangoService/Tango/volumes`? This actually complicates the deployment a little bit.

fanpu: The reason is that for non-Dockerfile setups this may be confusing for new users as it may seem to suggest that the Tango installation must be located in /opt/TangoService

fanpu: The reason is that for non-Dockerfile setups this may be confusing for new users as it may seem to suggest that the Tango installation must be located in /opt/TangoService

PR #175, opened 2020-10-08
Use env vars for config management

Huahan98:
- removed `config.py.docker`. Use `config.template.py` for docker deployment. 

How to test:
see https://github.com/autolab/docker/pull/7


config.template.py
```diff
@@ -18,7 +19,7 @@ class Config(object):
     # Unique prefix that defines VM name space for this Tango
     # version. When working in development, this prefix should be your
     # unique identifier. The "prod" prefix is reserved for production
-    PREFIX = "local"
+    PREFIX = "prod" if os.getenv("DOCKER_DEPLOYMENT").lower() == "true" else "local"
```
fanpu, line 22: This would break if the env var was not defined, as it becomes None.

I also think it would be more extensible if we allowed the DOCKER_DEPLOYMENT variable to directly specify the prefix, so that if the user wants to change it to development or test they can just do that easily by changing the docker-compose file

config.template.py
```diff
@@ -45,7 +46,7 @@ class Config(object):
     # Docker is being used as the VMMs.
     # It must be an absolute path with trailing slash, i.e
     # /opt/TangoService/Tango/volumes/
-    DOCKER_VOLUME_PATH = ""
+    DOCKER_VOLUME_PATH = "/opt/TangoService/Tango/volumes/" if os.getenv("DOCKER_DEPLOYMENT").lower() == "true" else ""
```
fanpu, line 49: Similar issue to previous comment

fanpu (changes requested):
Please see comments!

config.template.py
```diff
@@ -18,7 +19,7 @@ class Config(object):
     # Unique prefix that defines VM name space for this Tango
     # version. When working in development, this prefix should be your
     # unique identifier. The "prod" prefix is reserved for production
-    PREFIX = "local"
+    PREFIX = os.getenv("DOCKER_DEPLOYMENT", "local").lower()
```
fanpu, line 22: It would be easier for users to modify this line of config if the "local" part is separated out, such that it's the only thing on the RHS (similar to setting DOCKER_VOLUME_PATH)

fanpu (approved):

Issue #183, opened 2020-12-05
Better and more unit tests


fanpu:
The state of unit testing for Tango can be very much improved. Currently, we only have simple tests for `JobQueue`, and a boilerplate test file for `Preallocator`.

Other classes which should have unit tests added:
- `TangoJob`
- `Worker` (this might be tricky)
- `TangoREST`
- `TangoRemoteDictionary`
- `TangoNativeDictionary`
- `TangoRemoteQueue`

PR #187, opened 2020-12-10
Tangodict code improvements

### LINKED ISSUE #171: Code improvement opportunities
### Expected Behavior
N/A, same behavior, syntax change

### Actual Behavior
N/A, same behavior, syntax change

### Steps to Reproduce the Behavior
N/A, same behavior, syntax change

From this comment https://github.com/autolab/Tango/pull/160#issuecomment-674392496:

Most if not all of the keys() calls in the code do not require the list form of the keys (you only need to listify the keys if you either want to use [] to access elements of the key list, or if you will alter the dict while iterating over the keys). Furthermore, a more readable way to do
`if vmName in self.machines.keys():`

would be to implement `__contains__` in the Tango*Dictionary classes themselves and then call

`if vmName in self.machines:`

instead. TangoRemoteDictionary could use HEXISTS to implement `__contains__` and not have to fetch everything.

clairecw:
Implements __contains__ in the Tango*Dictionary classes, replacing any unnecessary calls to `.keys()`. 
Also adds tests for TangoDictionary accordingly. 

Resolves issue [171](https://github.com/autolab/Tango/issues/171).

### DIFF
--- a/tests/testObjects.py
+++ b/tests/testObjects.py
@@ -15,21 +15,47 @@ class TestObjects(unittest.TestCase):
             __db = redis.StrictRedis(
                 Config.REDIS_HOSTNAME, Config.REDIS_PORT, db=0)
             __db.flushall()
+        
+        self.test_entries = {
+            "key": "value",
+            0: "0_value",
+            123: 456,
+        }
 
+    def runDictionaryTests(self):
+        test_dict = TangoDictionary("test")
+        assert test_dict.keys() == []
+        assert test_dict.values() == []
+
+        for key in self.test_entries:
+            test_dict.set(key, self.test_entries[key])
+        
+        for key in self.test_entries:
+            assert key in test_dict
+            assert test_dict.get(key) == self.test_entries[key]
+
+        for (key, val) in test_dict.items():
+            assert self.test_entries.get(key) == val
+
+        assert test_dict.keys() == [str(key) for key in self.test_entries.keys()]
+        assert test_dict.values() == list(self.test_entries.values())
+        assert "key_not_present" not in test_dict
+        assert test_dict.get("key_not_present") is None
+
+        test_dict.set("key", "new_value")
+        assert test_dict.get("key") == "new_value"
+
+        test_dict.delete("key")
+        assert "key" not in test_dict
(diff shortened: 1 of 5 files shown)


fanpu:
Hi @clairecw , thank you for your PR! It looks really good in general, and I really appreciate the fact that there are unit tests. A couple of more complicated unit tests would also be helpful. Do you have other instructions for end-to-end testing that you could share as well?

clairecw:
Hi @fanpu, thanks so much for the feedback! Yes, I've added unit tests testing a bit more functionality for the dictionary object specifically - if you could TAL when you get the chance and see if there's anything else you had in mind that'd be awesome! 
Lmk also if you'd like to see unit tests for the other Tango objects.

As for E2E tests, I suppose there could be tests for each of the functions in TangoServer, and maybe another file testing the endpoints themselves.

fanpu:
Great job and thanks! I've looked through the changes and tests and they look good, and I've also tested it with Autolab.

One last change before it's ready to be merged: could you use the [unittest](https://docs.python.org/3/library/unittest.html) methods for asserting the tests instead of the native Python assert? You can refer to [this file](https://github.com/autolab/Tango/blob/master/tests/testJobQueue.py) for examples.

fanpu:
Tests for other Tango objects would also definitely be greatly appreciated, see https://github.com/autolab/Tango/issues/183 for some ideas

tests/testObjects.py
```diff
@@ -0,0 +1,61 @@
...
+        
+        self.test_entries = {
+            "key": "value",
+            0: "0_value",
+            123: 456,
+        }
+
+    def runDictionaryTests(self):
+        test_dict = TangoDictionary("test")
+        assert test_dict.keys() == []
```
fanpu, line 27: Change to unittest assert

fanpu (changes requested):

tests/testObjects.py
clairecw: This makes sense! Thanks!

fanpu (approved):
Excellent job and thanks once again for your contributions!

PR #190, opened 2020-12-12
Fixes for 'concurrency' issues in JobQueue and JobManager

### LINKED ISSUE #142: jobs with high ids stay waiting when the job id counter rolls over
It seems that tango always schedules jobs with lower ids first, so when the job id counter rolls over, any job that was WAITING when the new job 1 is enqueued doesn't run unless the queue drains completely.

### LINKED ISSUE #182: Tango concurrency and locking issues
There are potential concurrency issues in parts of the codebase that results in execution traces which causes Tango to behave anomalously.

See:
https://github.com/autolab/Tango/issues/138 (was replicable until very recently, but we used a workaround to fix it instead of resolving the root issue of why it was happening)

mojojojo99:
Fixes #182 and indirectly fixes #142

Changes proposed in this PR:
-  The most glaring issue fixed was [here](https://github.com/autolab/Tango/blob/43feeafb315ab7ac47a24818af95efacc6e8f0a5/jobManager.py#L103). I included a check for whether the pre-allocated VM in jobManager.py returns a `None` before attempting to do the logging. In the case where there are not enough VMs in the pool available, `None` can be returned. This previously caused an exception to be thrown and for the job to be considered Dead instead.
- Added a TangoQueue for unassigned jobs. This keeps a FIFO queue of live jobs that have been added and not yet assigned to a worker or vm. Whenever we `getNextPendingJob`, we simply just have to pop off the FIFO queue. Some implications:
       - This helps to indirectly avoid the possible starvation problem as mentioned in the issue #142, since jobs are assigned or run based on their order in the FIFO queue, rather than arbitrarily based on their jobIDs. 
       - I've also used a blocking call to the redis list `lpop`. This means that `getNextPendingJob` will block until there is a next job available in the job queue. Previously, it seems like we have the jobManager spinning when there are no jobs, which might be less desirable. 
      - To accommodate previous API methods in the jobManager like `delJob` etc., I have added methods to the `TangoQueue` object to add functions to remove items in the queue.  
- Removed some methods in the jobQueue which are not used currently and do not seem to be useful. I also changed the `getNextPendingJobReuse` method so that it will be less expensive. Previously, it was looping through the hash table to find a job and initialiaze a pool of VMs for it. However, this can be simplified if we simply retrieve the job and pass it to a helper function to do this initialization.
- Changed `initializeVM` to `createVM` in the `Worker`-- refer to the note below for more details. 
- Added more comments and docs to the functions within jobQueue. There were several assumptions (or preconditions required) in some methods and made them explicit in these comments. 
-  Added some minor error handling within jobQueue. I checked most of the functions to make sure that error codes are being handled. 
- Fixed some of the unit tests for jobQueue so that it aligns with the modified api.

#### QA
- More tests for this are in the making. We realize there aren't really any integrated tests present currently that makes sure that the `server` and the `jobManager` behaves as expected together, so we may be writing more of these. 

*Note:
- I have previously raised this up in the issue comments, but because of the exception thrown during the logging in jobManager, several branches and code in the `Worker` are not tested and used at all. This is the case when a worker is started with `preVM` is `None`. It seems like the current behavior of the worker will be to `initialize` a new VM whenever there is not already a pre-allocated VM for it. This might cause some issues and unexpected behavior for the case when `Config.REUSE_VMMS` is set to true -- when there is a flood of jobs to the autograder, many of these vms might be created in this worker. Because of the fact that they use the call `initializeVM` to immediately initialize and create a new vm instance, this vm is not added to the pool and is also untracked anywhere else. Based on my limited understanding, it seems like this vm will not be terminated. In the case where ec2ssh is used for example, this might be undesirable. I have replaced this call with `createVM` at least, so that the pool keeps track of this new vm. The preallocator will then manage the pools by terminating instances when there are too many later on. 

However, a more important question is, what is the desired behavior of the autograder when there are not enough VMs? Perhaps we could consider simply blocking the job there, and waiting for another vm to be freed instead of always creating new VMs. This will potentially cause a lot more VMs to be created during peak usage.

You might want to check out #191, though it is still a major WIP. I wrote some tests and it seems like these VMs are being created on the fly when there are no preallcoated vms.

### DIFF
--- a/jobManager.py
+++ b/jobManager.py
@@ -67,7 +67,7 @@ class JobManager(object):
         while True:
             # Gets the next pending job/ id
             # Blocks until we get a next job
-            job, _ = self.jobQueue.getNextPendingJobFromQueue()
+            job = self.jobQueue.getNextPendingJobFromQueue()
 
             if not job.accessKey and Config.REUSE_VMS:
                 vm = self.jobQueue.reuseVM(job)
--- a/jobManager.py
+++ b/jobManager.py
@@ -69,6 +69,8 @@ class JobManager(object):
             # NOTE: This grabs and acquires the lock on the queue
             # If there is no pending job, job and id is None
             job, _ = self.jobQueue.getNextPendingJobFromQueue()
+            # If there is no pending job, then we continue with 
+            # the loop
             if job is None:
                 continue
 
--- a/jobManager.py
+++ b/jobManager.py
@@ -67,7 +67,7 @@ class JobManager(object):
         while True:
             # Gets the next pending job/ id
             # Blocks until we get a next job
-            job = self.jobQueue.getNextPendingJobFromQueue()
+            job = self.jobQueue.getNextPendingJob()
 
             if not job.accessKey and Config.REUSE_VMS:
                 vm = self.jobQueue.reuseVM(job)
--- a/jobManager.py
+++ b/jobManager.py
@@ -96,19 +96,17 @@ class JobManager(object):
                         preVM = self.preallocator.allocVM(job.vm.name)
                     vmms = self.vmms[job.vm.vmms]  # Create new vmms object
 
-                if job is None:
-                    raise Exception("here1")
- 
-                # Now dispatch the job to a worker 
-                self.log.info("Dispatched job %s:%d to %s [try %d]" %
+                if (preVM is not None):
+                    # Log the preallocated vm only in the case where a vm 
+                    # was allocated
+                    self.log.info("Dispatched job %s:%d to %s [try %d]" %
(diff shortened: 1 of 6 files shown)


fanpu:
Thank you for your PR! There are a lot of critical issues being addressed by your changes. 

However, it is quite hard to review the entirety of this PR as many of them are making changes to core components and so I will need to be able to reason through each change completely before being comfortable with approving it.

Would it be possible to split this PR up into independent enhancements instead?

fanpu (changes requested):
Break up into smaller PRs

mojojojo99:
This has been split into PRs #195 #196 #197.

PR #194, opened 2020-12-15
Type annotations

CCCodes:
Fixes autolab#185.

Changes proposed in this PR:
Add type annotations for

jobManager.py
jobQueue.py
tangoObjects.py
worker.py
(Branched from tests)
Some type annotations are best guesses because it was hard to determine exactly what types were acceptable. I ran all unittests and new sampleTest on it and got the same results with / and without the annotations, so there were no problems in those cases.


fanpu (changes requested):
Hi @CCCodes, thanks for your PR!

I noticed that there are some changes unrelated to type annotations in the PR, would it be possible to only isolate the changes for type annotation as I assume that that is the intent of your PR?

In addition, there are also `env/*` files as well as pyenv.cfg being committed in this repo, which is not necessary. Could those be removed as well?

After that I will take a close look at the type annotations again. Thanks!

CCCodes:
Thanks, I made a new pull request with the changes!

PR #197, opened 2020-12-15
Slightly more inline comments

mojojojo99:
Changes proposed in this PR:
- Added more inline comments in `jobQueue`. I also made some assumptions in the code more explicit. 
- Return error code when `addDead` fails to get a `jobId` -- the current set up seems to ignore failures. Let me know if we actually want to ignore the error instead.


fanpu (approved):
LGTM, logic looks good and I agree that silently failing on error is bad, so thanks for fixing this!

PR #200, opened 2020-12-17
Type annotations fixed

### LINKED ISSUE #185: Type annotations
Now that Tango is on Python 3, it would be really useful to add [Python 3 typing hints](https://docs.python.org/3/library/typing.html) to the codebase. Using [mypy](http://mypy-lang.org/) to achieve this would be a good idea.

CCCodes:
Fixes #185.

Changes proposed in this PR:
Add type annotations for
- jobManager.py
- jobQueue.py
- tangoObjects.py
- worker.py
- tango.py
- preallocator.py

Some type annotations are best guesses because it was hard to determine exactly what types were acceptable. I ran all unittests on it and got the same results with / and without the annotations, so there were no problems in those cases.


fanpu:
Hi @CCCodes, thank you for updating the PR based on my previous comments! Apologies for the delay in response, I'll be reviewing it now

tango.py
```diff
@@ -209,7 +209,7 @@ def getInfo(self):
     #
     # Helper functions
     #
-    def resetTango(self, vmms):
+    def resetTango(self, vmms: list) -> None:
```
fanpu, line 212: `list` should be `List[<some type]` instead, and can be imported with `from typing import List`

jobQueue.py
```diff
@@ -313,7 +315,7 @@ def makeDead(self, id, reason):
         self.log.debug("makeDead| Released lock to job queue.")
         return status
 
-    def getInfo(self):
+    def getInfo(self) -> dict:
```
fanpu, line 318: `Dict` should be capitalized and contain the type mapping, see https://docs.python.org/3/library/typing.html#typing.Dict

jobQueue.py
```diff
@@ -353,7 +355,7 @@ def getNextPendingJob(self):
         self.log.debug("getNextPendingJob| Released lock to job queue.")
         return job
  
-    def reuseVM(self, job):
+    def reuseVM(self, job: TangoJob) -> Optional['vm']:
```
fanpu, line 358: 'vm' is probably not a valid type

tangoObjects.py
```diff
@@ -9,13 +9,14 @@
...
 import pickle
+from typing import Optional, Iterator, Tuple
 from queue import Queue
 from config import Config
 
 redisConnection = None
 
 
-def getRedisConnection():
+def getRedisConnection() -> redis.StrictRedis:
```
fanpu, line 19: Figuring out the type of this is tricky, it's actually `redis.client.Redis`. You can do it by installing [mypy](http://mypy-lang.org/), creating a test python file (say test.py), adding the following stub:

```
import redis
reveal_type(redis.StrictRedis(host='localhost', port=6379, db=0))
```

`reveal_type` is a feature from `mypy` that has an example here: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated

tangoObjects.py
```diff
@@ -257,31 +258,31 @@ def get(self, block=True, timeout=None):
...
 
-    def _clean(self):
+    def _clean(self) -> None:
         self.__db.delete(self.key)
 
 # This is an abstract class that decides on
 # if we should initiate a TangoRemoteDictionary or TangoNativeDictionary
 # Since there are no abstract classes in Python, we use a simple method
-def TangoDictionary(object_name):
+def TangoDictionary(object_name) -> 'TangoDictionary':
```
fanpu, line 285: This should be `tangoObjects.TangoRemoteDictionary`

tangoObjects.py
```diff
@@ -306,68 +307,68 @@ def set(self, id, obj):
...
         else:
             return None
 
-    def keys(self):
+    def keys(self) -> list:
         keys = map(lambda key : key.decode(), self.r.hkeys(self.hash_name))
         return list(keys)
 
-    def values(self):
+    def values(self) -> list:
```
fanpu, line 322: Same comment on list as before

fanpu (changes requested):
Great job overall and thanks so much for your contributions! There are some compound types and custom datatypes whose annotations are wrong but it's also partially the fault of the lack of a CI pipeline in Tango which means that you won't really be able to check if they are actually working. Could you address the comments, and once those are done it should be good to go!

CCCodes:
Thanks for the feedback, I will work on the changes now!

fanpu (approved):
Thanks for making the changes @CCCodes! I looked through the changes again and found a few more issues but I'll be able to take care of them. Merging changes into a temporary branch first in order to fix these issues. Once again thanks for your contributions!

PR #202, opened 2021-01-10
Add module system

fanpu:
This adds modules to Tango, as the lack of modules currently prevents third party tools  (i.e [MonkeyType](https://github.com/Instagram/MonkeyType), and other testing frameworks) from effectively interfacing with Tango due to import issues as the absence of modules means that the working directory will not be included in the Python search path.

- Rename `restful-tango` to `restful_tango` as dashes are not accepted in the module system
- Reorder imports
- Formatting improvements
- Change usage of importing * to specify which objects are imported

Testing plan:
- Ran jobs on Autolab and verified that it was successfully autograded


Huahan98 (approved):
Tested grading assignments on docker-compose installation. LGTM!

PR #205, opened 2021-01-30
Fix buffer overflow bug in autodriver

fanpu:
Fixes the issue where if the file is BUF_SIZE the buffer will not be properly null terminated, and also buffer allocated was too small to hold the null terminator.

This was inspired by https://github.com/autolab/Tango/commit/c1aa0ce4139b6e4e7aa119d18323ffde3a650221 but because of how much the code has diverged cherry picking is not possible.


Huahan98 (approved):

cg2v:
FWIW, this seems unnecessary, there's no strlen() in this function, and write does not care about NULLs (the parameter to write is no more than the return from the read)

PR #206, opened 2021-02-18
Code formatting fixes from Black

oliverli:
[Black](https://github.com/psf/black) is a Python code linter. Adds automated formatting fixes from Black with the end goal of running all our code through PyLint strictly


oliverli:
Closed because  #218 is more updated

PR #208, opened 2021-03-15
Allow tango-cli to use SSL

cg2v:
Give tango-cli.py a switch to enable SSL communication with tango

Changes proposed in this PR:
- Make the "protocol" (http) in tango urls passed to requests.get a variable instead of a constant in all the urls
- Add a switch -S (--ssl) to switch the protocol to https. The switch also changes the default port from 3000 to 443

### DIFF
--- a/clients/tango-cli.py
+++ b/clients/tango-cli.py
@@ -465,6 +465,9 @@ if args.ssl:
     _tango_protocol = 'https'
     if args.port == 3000:
         args.port = 443
+    else:
+        print('SSL option ignored as custom port specified')
+
 
 try:
     response = requests.get('%s://%s:%d/' % (_tango_protocol, args.server, args.port))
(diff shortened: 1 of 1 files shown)


clients/tango-cli.py
```diff
@@ -453,8 +461,14 @@ def router():
     parser.print_help()
     sys.exit(0)
 
+if args.ssl:
+    _tango_protocol = 'https'
+    if args.port == 3000:
+        args.port = 443
+
```
fanpu, line 468: It would be good to print a warning here for the else branch that the SSL option is not being respected because another port is already being specified

fanpu (approved):
Other than the nit for the logging, everything else looks pretty sound. Thank you!

fanpu (approved):

cg2v:
The option isn't actually ignored. It's only the port override that doesn't happen.

The idea is that the *default* port is changed by `--ssl`, but if the caller specified a non-default port, that port is still used.

fanpu:
Thanks for the explanation, that makes sense, made #212 to remove that comment.

PR #209, opened 2021-03-15
Stop using the python-future module

cg2v:
Since this branch is now python3 only, the future module is not needed

Changes proposed in this PR:
-  Remove from __future__ import print_function
-  remove all imports of future and builtins
-  remove future from requirements.txt

python-future only has effects on python2. Since this branch is now supposedly python3 only, it should not be needed.
Also note that the previous code in tango-cli.py would not have worked on python2, because new style urllib imports were used before `standard_library.install_aliases()` was called

### DIFF
--- a/clients/tango-cli.py
+++ b/clients/tango-cli.py
@@ -4,7 +4,6 @@
 # tango-cli.py - Command line client for the RESTful Tango.
 #
 
-from __future__ import print_function
 import urllib.error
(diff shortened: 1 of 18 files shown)


fanpu (approved):
Modifications LGTM, tested with Autolab autograding submission as well just to make sure and everything seems fine

PR #211, opened 2021-03-23
Add linting pre-commit hook, and do linting

oliverli:
Installed the [`pre-commit`](https://pre-commit.com/) package, and used it to create a pre-commit hook that runs some of its own linting fixes (lines 7-10 in `.pre-commit-config.yaml`) and Black.

Created a new `requirements-dev.txt` with the two development dependencies.

Ran the two linters. Black initially had some trouble with `vmms/tashiSSH.py` and `vmms/ec2SSH.py` due to a bug in Black, which was bypassed by first running Black with the `--fast` flag.


xinyis991105 (approved):
LGTM! I saw that the hooks for pre-commit are all available in v3.2.0 for pre-commit but is there any particular reason why we use v3.2.0 over v3.4.0?

xinyis991105 (approved):
LGTM!

PR #212, opened 2021-03-24
Revert warning on custom port with SSL option

fanpu:
Removes the warning when using SSL but also specifying SSL port, as it can be part of a legitimate workflow

https://github.com/autolab/Tango/pull/208#issuecomment-805878811


victorhuangwq (approved):
LGTM

PR #216, opened 2021-10-06
fix: catch the network error in docker build

wongsingfo:
Fixes the issue where if a network error occurs, the docker build process will not be terminated properly. By default, pipeline (`curl -sSL https://get.docker.com/ | sh`) doesn't produce a failure return code when the last command (`sh` here) returns zero, even if `curl` returns non-zero exit status.


fanpu (approved):
Sorry about the delay, but this looks good. Thanks for your contribution!

PR #217, opened 2021-11-11
Allow failed posting of autograding results in Tango to be retried

xinyis991105:
This is based on a customized fix from Chaskiel. See https://github.com/cg2v/Tango/commit/885b9649a1021be9c57748c3a49ff771a8a2c04b.

@cg2v Thanks for suggesting the change! Let us know if you think it's ok for us to adopt it. We will also do further testing to make sure Tango functions normally before we merge it in.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,6 @@ boto==2.49.0 # used only by ec2SSH.py
 plumbum==1.6.9
 pyflakes==2.1.1
 redis==3.4.1
-requests==2.23.0
+requests==2.26.0
 rpyc==4.1.4
 tornado==4.5.3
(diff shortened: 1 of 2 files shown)


cg2v:
I think the summary here is confusing. What's being retried is posting the autograding results back to the frontend, not a failed job itself. 

That said, there's no external dependencies to worry about, and this should not have any effect unless the initial POST of the autograde_done callback fails.

xinyis991105:
Thanks for the clarification!

xinyis991105:
We need to upgrade the `requests` library to make the `Retry` method work.
Tested by:
1. Single submission grading.
2. Bulk submission grading.

victorhuangwq (approved):
LGTM

PR #218, opened 2022-02-06
Integrate pre-commit linting and lint Python files

oliverli:
Uses [pre-commit](https://pre-commit.com/) and its [generic linters](https://github.com/pre-commit/pre-commit-hooks), as well as [Black](https://github.com/psf/black) for Python specific linting


victorhuangwq (approved):
Works well! 
Tried making a commit with issues, and black was ran as part of the commit.

PR #219, opened 2022-02-06
Runs pre-commit as part of CI

victorhuangwq:
This prevents users from overriding / not using pre-commit


fanpu:
Tested using https://github.com/autolab/Tango/pull/220

victorhuangwq:
@fanpu

fanpu (approved):
Tested again using https://github.com/autolab/Tango/pull/220, with expected results. LGTM!

PR #225, opened 2022-10-31
Add partial output support for distDocker

victorhuangwq:
Adds partial output support for distDocker, since CMU's Autolab uses distDocker

### DIFF
--- a/vmms/distDocker.py
+++ b/vmms/distDocker.py
@@ -475,3 +475,39 @@ class DistDocker(object):
                 result.add(re.sub(r".*/([^/]*)", r"\1", row_l[0]))
 
         return list(result)
+
+    def getPartialOutput(self, vm):
+        """getPartialOutput - Get the partial output of a job.
+        It does not check if the docker container exists before executing
+        as the command will not fail even if the container does not exist.
+        Gets the first MAX_OUTPUT_FILE_SIZE bytes of the feedback file
+        """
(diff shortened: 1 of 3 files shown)


vmms/distDocker.py
```diff
@@ -475,3 +475,39 @@ def getImages(self):
...
+                self.log.debug("Lost persistent SSH connection")
+                return ret
+
+        cmd = "(docker exec -it %s head -c %s autograde/output.log)" % (
+            instanceName,
+            config.Config.MAX_OUTPUT_FILE_SIZE,
+        )
+
+        output = subprocess.check_output(
+            ["ssh"]
```
fanpu, line 506: Safer to pass in the args as a list to avoid command injection (might be possible via vm.domain_name via an adversarially crafted call to Tango, i didn't trace the code fully though)

See https://docs.python.org/3/library/subprocess.html#frequently-used-arguments second paragraph about `args`

victorhuangwq: If you are referring to this string `"%s@%s" % (self.hostUser, vm.domain_name)`, it isn't passed in as args as list because it is needed to be a singular string of host@domain. 

The rest of the args are list of strings. `DistDocker._SSH_FLAGS` is a list of strings

damianhxy (approved):
code looks good

lgtm!

PR #226, opened 2022-12-05
Remove interactive flag from exec

victorhuangwq:
Fixes the issue where getPartialOutput does not work in Autolab Docker. 

Achieves it by removing the `-it` interactive flag that is meant for an interactive command, which throws an error when run in python in Docker.

**How to test**
Spin up Autolab Docker, do an autograding of hello.c with this content

```
/* Solution for the Hello Lab */
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main()
{
    for (int i = 0; i < 5; i++)
    {
        /* code */

        printf("Hello, world\n");
        sleep(10);
        fflush(stdout);
    }
    return 0; /* important to return zero here */
}
```
Partial output should work (as long as the error of not appearing on the queue does not come up). Without this PR, it will always not work.

### DIFF
--- a/vmms/distDocker.py
+++ b/vmms/distDocker.py
@@ -497,7 +497,7 @@ class DistDocker(object):
                 self.log.debug("Lost persistent SSH connection")
                 return ret
 
-        cmd = "(docker exec -it %s head -c %s autograde/output.log)" % (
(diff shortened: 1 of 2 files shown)


20wildmanj (approved):
LGTM

PR #227, opened 2022-12-05
[WIP] Hotfix for job-queue

20wildmanj:
## Description

Fixes an issue where jobs' assignment state (`assigned`) field are not consistent nor updated properly, resulting in a condition where a job is live, and is assigned and running on a VM, but the job in the `liveJobs` dictionary has `assigned==False`.

## Testing (local and docker-compose)

- Make multiple submissions of long running submissions
- check to see that partial output gets displayed and that the queue for submissions works

### DIFF
--- a/jobManager.py
+++ b/jobManager.py
@@ -73,8 +73,7 @@ class JobManager(object):
                     time.sleep(Config.DISPATCH_PERIOD)
 
             try:
-                # Mark the job assigned
-                job.makeAssigned()
+
                 # if the job has specified an account
                 # create an VM on the account and run on that instance
                 if job.accessKeyId:
@@ -108,7 +107,8 @@ class JobManager(object):
                     "%s|Dispatched job %s:%d [try %d]"
                     % (datetime.utcnow().ctime(), job.name, job.id, job.retries)
                 )
-
+                # Mark the job assigned
+                self.jobQueue.assignJob(job.id, preVM)
                 Worker(job, vmms, self.jobQueue, self.preallocator, preVM).start()
 
             except Exception as err:
--- a/tests/testJobQueue.py
+++ b/tests/testJobQueue.py
@@ -93,11 +93,11 @@ class TestJobQueue(unittest.TestCase):
         self.assertEqual(str(ret_job_2.id), self.jobId2)
 
     def test_getNextPendingJob(self):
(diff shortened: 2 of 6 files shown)


jobManager.py
```diff
@@ -108,7 +107,8 @@ def __manage(self):
                     "%s|Dispatched job %s:%d [try %d]"
                     % (datetime.utcnow().ctime(), job.name, job.id, job.retries)
                 )
-
+                # Mark the job assigned
+                self.jobQueue.assignJob(job.id, preVM)
```
michellexliu, line 111:
```
if preVM is None:
    job.makeAssigned()
else:
    self.jobQueue.assignJob(job.id, preVM)
```
makes this work locally for me... still can't get autolab-docker to work so I haven't been able to test it there though

20wildmanj: ah but this makes it work on docker lol, which is so confusing because they essentially run the same code but in different places...

20wildmanj: Actually nvm I haven't tried this with the edited comment. Let me see what happens

tests/testJobQueue.py
```diff
@@ -93,11 +93,11 @@ def test_get(self):
         self.assertEqual(str(ret_job_2.id), self.jobId2)
 
     def test_getNextPendingJob(self):
-        self.jobQueue.assignJob(self.jobId2)
+        self.jobQueue.assignJob(self.jobId2, None)
```
michellexliu, line 96: nit: change the header of `assignJob` to `def assignJob(self, jobId, vm = None)` so that you don't have to manually pass in None everytime you call it.

worker.py
```diff
@@ -198,8 +197,7 @@ def run(self):
             # Assigning job to a new VM
             else:
                 self.log.debug("Assigning job to a new VM")
-                self.job.vm.id = self.job.id
-                self.job.updateRemote()
+                self.job.makeVM(self.job.id)
```
michellexliu, line 200: since we're setting `self.vm` instead of `self.vm.id` in `makeVM`, wouldn't this cause VM be set to a string here?

20wildmanj: You're right, it should be `self.job`

michellexliu (approved):
lgtm!

PR #228, opened 2022-12-08
[hotfix] Check element exists before removing for Python thread-safe queue

20wildmanj:
## Description
Fixes issue found in Tango deployments that use the default Python thread-safe queue instead of `TangoRemoteQueue`, where it was possible for `remove` to be called to remove an element already  removed from a queue, causing an unhandled exception. 

image

This hotfix merely will check to see if the element exists in the queue before calling the `remove` method, which is consistent with `TangoRemoteQueue`, which does not raise an exception in the same scenario as well.

## Testing
- set `USE_REDIS=False` in `config.py`
- Run `python -m unittest tests/testJobQueue.py`, see all tests run successfully
- run only restful tango
- try uploading files to autograded assignment, see that they get graded properly

### DIFF
--- a/jobQueue.py
+++ b/jobQueue.py
@@ -253,11 +253,10 @@ class JobQueue(object):
 
         job = self.liveJobs.get(jobId)
 
-        # Remove the current job from the queue
-        self.unassignedJobs.remove(int(jobId))
 
         self.log.debug("assignJob| Retrieved job.")
         self.log.info("assignJob|Assigning job ID: %s" % str(job.id))
+        
         job.makeAssigned()
         job.makeVM(vm)
 
--- a/tangoObjects.py
+++ b/tangoObjects.py
@@ -230,7 +230,8 @@ class ExtendedQueue(Queue):
 
     def remove(self, value):
         with self.mutex:
-            self.queue.remove(value)
+            if value in self.queue:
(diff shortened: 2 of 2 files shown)


michellexliu (approved):
verified that it works; lgtm

cg2v:
While this will certainly prevent exceptions, is it actually the right fix? This looks like another variant of #207

20wildmanj:
Yes we have looked at #207, though because my exposure to Tango has only been through hotfixes the past week I'm not 100% in terms of understanding the states of Tango. There's definitely a better solution, but making it happen right now may be difficult in terms of time.

cg2v:
I'm also concerned because the exception I encountered was in assignJob, which only has one callsite. How was jobQueue.assignJob called on a job that was not in the unassigned queue?

michellexliu:
The jobQueue.assignJob replaced job.makeAssigned() in __manage in #227 to fix a bug where the vm was never getting set, leading to issues in our autolab-docker installation. To be honest, I'm not too familiar with the codebase and am not sure when the different states and such get set / how it would get called on a job in the unassigned queue, so we would need to look more into this in the future.

20wildmanj:
Closing this PR for now, will return to this issue with a better fix later.

PR #231, opened 2023-04-20
Add Redis version

jessehartloff:
The old link used to download redis (http://download.redis.io/releases/redis-stable.tar.gz) was removed and caused a 404. Updated to a specific version so redis will still be installed (http://download.redis.io/releases/redis-3.2.6.tar.gz).

Version 3.2.6 is old. This version was chosen because it's what's running on the AutoLab deployment so it's proven to work. Latest version of redis is 7.0.9 at the time of writing.

Changes proposed in this PR:
-  Fix redis download link in Dockerfile


jessehartloff:
Sorry, wrong repo

PR #234, opened 2023-05-17
Update Dockerfile to use Ubuntu 20.04, Update dependencies

20wildmanj:
#230 updated redis from 3.4.1 to 4.4.4, however Python3 on Ubuntu 18.04 has a version of 3.6.x while `redis-py` requires a version >= 3.7. Ubuntu 20.04 comes with Python 3.8, so this PR bumps up Ubuntu in order to keep the Tango container for Docker working.

Also updated Ubuntu libraries to be compatible with 20.04:
- `tcl` from 8.5 to 8.6
- `libgcrypt-dev` from 11 to 20

Verified by building container on Nightly, and running an autograded job and ensuring job completed successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Start with empty ubuntu machine
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 MAINTAINER Autolab Development Team "autolab-dev@andrew.cmu.edu"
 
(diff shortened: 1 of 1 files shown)


damianhxy (approved):
Successfully built container, as well as (re)grade a submission

LGTM

PR #237, opened 2023-06-03
Refactor server.py to be compatible with Tornado v6+

20wildmanj:
Tornado's `web.asynchronous` attribute was [deprecated in Tornado v5.x and removed in x6.x](https://stackoverflow.com/questions/55475879/tornado-v6-seems-to-have-dropped-tornado-web-asynchronous-coroutine-any-differe), making a refactor to the `server.py` code necessary. `server.py` now uses asynchronous and code that uses native coroutines (e.g. `async def`, `await`). Some RequestHandlers were also made synchronous since they just do simple fetches of data.

I made the refactor mainly by going through the [most recent Tornado user guide](https://www.tornadoweb.org/en/stable/guide/structure.html) as well as reading through some [example code](https://github.com/tornadoweb/tornado/blob/stable/demos/chat/chatdemo.py).

- The unblock function was completely removed, this seems to have been from a [code snippet](https://gist.github.com/lbolla/4594879#file-futures_test-py) that is now over 9 years old, and this Tornado code in general originates from the [initial commit](https://github.com/autolab/Tango/blob/537ad47b77b449b53f24c5e0a744c2f885c05c09/restful-tango/server.py) for this repo.

Testing:
Not the most comprehensive testing, but enough to check that nothing seems to break

- Tested that `restful_tango/server.py` runs on Python 3.7, 3.8, 3.9, and 3.10
- On Python 3.10:
  - Tested that autograding works, autograding results get returned
  - Tested that partial output works, queue information shows up
  - Tested that `getjob` displays correct information for job
  - Tested that `jobs` displays correct info
  - Tested that `jobs/tango_status` works
  - Tested that regrading works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Start with empty ubuntu machine
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 MAINTAINER Autolab Development Team "autolab-dev@andrew.cmu.edu"
 
@@ -24,9 +24,9 @@ RUN apt-get update && apt-get install -y \
 	python3 \
 	python3-pip \
 	build-essential \
-	tcl8.5 \
+	tcl8.6 \
 	wget \
-	libgcrypt11-dev \
+	libgcrypt20-dev \
 	zlib1g-dev \
 	apt-transport-https \
 	ca-certificates \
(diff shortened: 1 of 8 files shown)


damianhxy:
Tornado v6.3+ no longer supports Python 3.7

Source: https://www.tornadoweb.org/en/stable/releases/v6.3.0.html

damianhxy (approved):
Followed testing plan successfully

Discussed offline re: Python 3.7

LGTM

PR #240, opened 2023-09-17
Fix installation link in README.md to point to correct location

jlge:
## Description
The link pointing to the Tango installation instructions currently points to a page that doesn't exist.

## Changes
- Edited the link to reflect to the correct page


damianhxy (approved):
Changed my life 10/10 LGTM

PR #241, opened 2023-12-01
Add docker resource limits for jobs

evanyeyeye:
Adds the following resource limits to new jobs:
- Number of CPU cores
- Memory (in MB)

Addresses: @139

Resource limits can be verified using `docker stats` and `htop`.

### DIFF
--- a/restful_tango/tangoREST.py
+++ b/restful_tango/tangoREST.py
@@ -113,8 +113,8 @@ class TangoREST(object):
 
     def createTangoMachine(self, image, vmms=Config.VMMS_NAME, vmObj=None):
         """createTangoMachine - Creates a tango machine object from image"""
-        cores = Config.DOCKER_CORES_LIMIT
-        memory = Config.DOCKER_MEMORY_LIMIT
+        cores = getattr(Config, "DOCKER_CORES_LIMIT", None)
+        memory = getattr(Config, "DOCKER_MEMORY_LIMIT", None)
         if vmObj and "cores" in vmObj and "memory" in vmObj:
             cores = vmObj["cores"]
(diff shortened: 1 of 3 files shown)


restful_tango/tangoREST.py
```diff
@@ -111,16 +111,19 @@ def checkFileExists(self, directory, filename, fileMD5):
                 except IOError:
                     continue
 
-    def createTangoMachine(
-        self, image, vmms=Config.VMMS_NAME, vmObj={"cores": 1, "memory": 512}
-    ):
+    def createTangoMachine(self, image, vmms=Config.VMMS_NAME, vmObj=None):
         """createTangoMachine - Creates a tango machine object from image"""
+        cores = Config.DOCKER_CORES_LIMIT
+        memory = Config.DOCKER_MEMORY_LIMIT
```
damianhxy, line 117: These two lines should use `getattr` so that existing installations don't break when their Config object does not define the limits

Screenshot 2023-12-30 at 02 40 36

e.g.
```
cores = getattr(Config, "DOCKER_CORES_LIMIT", None)
memory = getattr(Config, "DOCKER_MEMORY_LIMIT", None)
```

damianhxy (changes requested):
Verified that limits are set (1GB memory, 1 CPU):

Screenshot 2023-12-30 at 02 57 44

(Ref https://phoenixnap.com/kb/docker-memory-and-cpu-limit)

However, need to avoid breaking existing installations -- see comment about `getattr`

damianhxy (approved):
Verified that autograding works when `Config.py` is outdated

LGTM

PR #267, opened 2024-12-13
Change how docker run args are built

cg2v:
There are 2 changes here
- Redoes distDocker:runJob to look more like localDocker:runJob. I did this so I could stop maintaining a patch that manually inserts a --memory switch in distDocker
- makes localDocker's runJob use list.append instead of + to build up the list. When I copied the code, I wanted to make it more pythonic

This isn't tested yet. I'll try to do that by next tuesday or so.

### DIFF
--- a/vmms/localDocker.py
+++ b/vmms/localDocker.py
@@ -167,13 +167,10 @@ class LocalDocker(object):
         args.extend(("sh", "-c"))
 
         autodriverCmd = (
-            "autodriver -u %d -f %d -t %d -o %d autolab > output/feedback 2>&1"
-            % (
-                config.Config.VM_ULIMIT_USER_PROC,
-                config.Config.VM_ULIMIT_FILE_SIZE,
-                runTimeout,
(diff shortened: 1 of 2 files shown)


cg2v:
tested with distDocker

vmms/localDocker.py
```diff
@@ -156,31 +156,28 @@ def runJob(self, vm, runTimeout, maxOutputFileSize, disableNetwork):
...
             )
         args = ["docker", "run", "--name", instanceName, "-v"]
-        args = args + ["%s:%s" % (volumePath, "/home/mount")]
+        args.append("%s:%s" % (volumePath, "/home/mount"))
         if vm.cores:
-            args = args + [f"--cpus={vm.cores}"]
+            args.append(f"--cpus={vm.cores}")
         if vm.memory:
-            args = args + ["-m", f"{vm.memory}m"]
+            args.append(f"--memory{vm.memory}m")
```
anthony-yip, line 163: need to add = between --memory

anthony-yip (changes requested):

PR #270, opened 2025-01-26
Copy in

KesterTan:
Attempts to fix the copy-in issue that we're still facing even after retries.

1. Captures invalid inputfiles structure 
2. Adds -p flag and permissions when creating autolab directory. This ensures that the specified directory and any necessary parent directories are created. If the directory already exists, the command does not raise an error, it would originally throw an error. 
3. Added thread pool to limit scp commands so that instance is not overloaded
4. Added log statements with job id for more in-depth logging.

Ran a thousand jobs on https://dev.autolabproject.com/courses/test-course/jobs?id=500 without failure.


clients/tango-cli.py
```diff
@@ -127,7 +131,8 @@
 parser.add_argument("--accessKeyId", default="", help="AWS account access key ID")
 parser.add_argument("--accessKey", default="", help="AWS account access key content")
 parser.add_argument("--instanceType", default="", help="AWS EC2 instance type")
-
+parser.add_argument("--ec2", action="store_true", help="Enable ec2SSH VMMS")
```
anthony-yip: can't you just check for whether config.VMMS_NAME == "ec2SSH"?

clients/tango-cli.py
```diff
@@ -257,10 +262,15 @@ def tango_addJob():
         if args.notifyURL:
             requestObj["notifyURL"] = args.notifyURL
 
+        if args.callbackURL:
```
anthony-yip: use .get()?

clients/tango-cli.py
```diff
@@ -257,10 +262,15 @@ def tango_addJob():
...
+            requestObj["callback_url"] = args.callbackURL
+
         requestObj["accessKeyId"] = args.accessKeyId
         requestObj["accessKey"] = args.accessKey
         requestObj["disable_network"] = args.disableNetwork
         requestObj["instanceType"] = args.instanceType
+        requestObj["ec2Vmms"] = args.ec2
+        requestObj["stopBefore"] = args.stopBefore
```
anthony-yip: requestObj should be a dataclass

jobManager.py
```diff
@@ -133,6 +141,13 @@ def __manage(self):
         tango.resetTango(tango.preallocator.vmms)
         for key in tango.preallocator.machines.keys():
             tango.preallocator.machines.set(key, [[], TangoQueue(key)])
+
+            # The above call sets the total pool empty.  But the free pool which
+            # is a queue in redis, may not be empty.  When the job manager restarts,
+            # resetting the free queue using the key doesn't change its content.
+            # Therefore we empty the queue, thus the free pool, to keep it consistent
+            # with the total pool.
+            tango.preallocator.machines.get(key)[1].make_empty()
```
anthony-yip, line 150: Wanted to draw attention that this makes sense only because either both the dictionary and the queue are local (so you can modify in-place), or are both Redis (so modifying a transient copy generated by [1] modifies the redis object as they share the same hash). 

For example, tango.preallocator.machines.get(key)[0].append(None) does nothing if we are using Redis.

preallocator.py
```diff
@@ -47,6 +47,7 @@ def update(self, vm, num):
         self.lock.acquire()
         if vm.name not in self.machines:
             self.machines.set(vm.name, [[], TangoQueue(vm.name)])
+            self.machines.get(vm.name)[1].make_empty()
```
anthony-yip, line 50: same comment as above regarding remote data structures

restful_tango/tangoREST.py
```diff
@@ -166,7 +166,11 @@ def convertJobObj(self, dirName, jobObj):
 
         ec2_vmms = False
         if "ec2Vmms" in jobObj:
-            ec2_vmms = True
+            ec2_vmms = jobObj["ec2Vmms"]
+
+        stopBefore = ""
+        if "stopBefore" in jobObj:
+            stopBefore = jobObj["stopBefore"]
```
anthony-yip, line 171: .get(), dataclass

worker.py
```diff
@@ -152,7 +152,7 @@ def notifyServer(self, job):
         except Exception as e:
             self.log.debug("Error in notifyServer: %s" % str(e))
 
-    def afterJobExecution(self, hdrfile, msg, returnVM):
+    def afterJobExecution(self, hdrfile, msg, returnVM, killVM=True):
```
anthony-yip: I would prefer detachVM over killVM, as "killVM" can still mean returning the VM to the free list

worker.py
```diff
@@ -161,7 +161,8 @@ def afterJobExecution(self, hdrfile, msg, returnVM):
         os.chmod(self.job.outputFile, 0o644)
 
         # Thread exit after termination
-        self.detachVM(return_vm=returnVM)
+        if killVM:
```
anthony-yip: Add a comment that explains why you wouldn't want to killVM (to SSH into the autograding VM to debug it)

vmms/ec2SSH.py
```diff
@@ -510,18 +513,21 @@ def waitVM(self, vm, max_secs):
             # Sleep a bit before trying again
             time.sleep(config.Config.TIMER_POLL_INTERVAL)
 
-    def copyIn(self, vm, inputFiles):
+    def copyIn(self, vm, inputFiles, job_id=None):
```
anthony-yip, line 516: Add a comment that job_id is only for debugging currently

anthony-yip (commented):
Left some comments, but a lot are just nitpicks. Talk to me how you want to proceed. I've also fixed some of these in my own PRs. Also, you mentioned thread pool but I'm not sure which one you're referring to?

worker.py
anthony-yip: detachVM: "The worker must always call this function before returning" which will not be the case if using stopBefore.

consider setting vm.keep_for_debugging=True and still call detachVM?

worker.py
```diff
@@ -254,6 +255,11 @@ def run(self):
                 )
             )
             self.log.debug("Waiting for VM")
+            if self.job.stopBefore == "waitvm":
+                msg = "Execution stopped before %s" % self.job.stopBefore
+                returnVM = True
```
anthony-yip, line 259: why set returnVM to True when the killVM=False makes the returnVM argument meaningless?

tangoObjects.py
```diff
@@ -319,6 +321,12 @@ def remove(self, item):
     def _clean(self):
         self.__db.delete(self.key)
 
+    def make_empty(self):
```
dwang3851, line 324: can we not do self.__db.delete(self.key) here instead? How is this function different from _clean? I'm not super familiar with Redis but it seems wrong that we have to iterate here

jobManager.py
```diff
@@ -133,6 +141,13 @@ def __manage(self):
...
         jobs = JobManager(tango.jobQueue)
 
         print("Starting the stand-alone Tango JobManager")
```
dwang3851: nit: log statement

dwang3851 (changes requested):
Overall looks good, left some small comments

clients/tango-cli.py
KesterTan: True, I changed it to depend on an earlier argument for vmms

jobManager.py
KesterTan: ok, changed

jobManager.py
KesterTan: It's almost always be the case that both are redis.

preallocator.py
KesterTan: Noted

restful_tango/tangoREST.py
KesterTan: Let's resolve this in a different PR

worker.py
KesterTan: Makes sense, remove detached vm

worker.py
KesterTan: Not needed because of the above comment

worker.py
KesterTan: Ok, I think it makes more sense now that killVM is removed

vmms/ec2SSH.py
KesterTan: Added

tangoObjects.py
KesterTan: We could, I think it was written this way to potentially do some sort of logging/processing but was subsequently removed

clients/tango-cli.py
```diff
@@ -252,41 +285,43 @@ def tango_addJob():
...
+        if get_arg('callbackURL'):
+            requestObj["callback_url"] = get_arg('callbackURL')
 
-        requestObj["accessKeyId"] = args.accessKeyId
-        requestObj["accessKey"] = args.accessKey
-        requestObj["disable_network"] = args.disableNetwork
-        requestObj["instanceType"] = args.instanceType
-        requestObj["ec2Vmms"] = args.ec2
-        requestObj["stopBefore"] = args.stopBefore
+        requestObj["accessKeyId"] = get_arg('accessKeyId')
```
anthony-yip: I'm pretty sure (I tested it) this syntax doesn't work now because requestObj is a dataclass. You have to initialize them in the constructor.

worker.py
```diff
@@ -161,8 +161,7 @@ def afterJobExecution(self, hdrfile, msg, returnVM, killVM=True):
         os.chmod(self.job.outputFile, 0o644)
 
         # Thread exit after termination
-        if killVM:
-            self.detachVM(return_vm=returnVM)
+        self.detachVM(return_vm=returnVM)
```
anthony-yip, line 164: new nit: this does not allow you to call detachVM with return_vm = true and replace_vm = true - feel free to ignore as this is irrelevant for ec2 which overrides the setting and i've fixed it in the spot instances pr anyway

anthony-yip (commented):
left 2 more comments, the ones I left last time were addressed

clients/tango-cli.py
KesterTan: Fixed.

clients/tango-cli.py
```diff
@@ -12,10 +12,43 @@
 import argparse
 import sys
 import os
+from dataclasses import dataclass, asdict
+from typing import Optional
 
 sys.path.append("/usr/lib/python2.7/site-packages/")
 
 
+def get_arg(name, default=None):
```
anthony-yip, line 21: nit: isn't this only different in the case of developer error, i.e. passing the wrong identifier as name or args.<here>, in which it changes an AttributeError to the standard error message that we wrote

anthony-yip (approved):

clients/tango-cli.py
KesterTan: I added the error handling back

dwang3851 (approved):

clients/tango-cli.py
KesterTan: done

clients/tango-cli.py
KesterTan: done

worker.py
KesterTan: ok, ignoring for now

Issue #279, opened 2025-03-21
Force job.retries to be an integer


anthony-yip:
![Image](https://github.com/user-attachments/assets/b8f4092d-7986-40f0-87f8-6674225fdd48)

Currently, this code implies that job.retries can be None. However, currently there is no way for that to be the case, and in the code we're basically treating None like -1. I think it would be appropriate to remove this guard and use type-checking to ensure that job.retries is always an integer.

Issue #277, opened 2025-03-17
Bug: in ec2SSH, if waitVM fails, the worker will keep retrying forever, irrespective of max_retries in the config


anthony-yip:
For example, waitVM might fail due to a ping fail or an ssh fail. The worker calls rescheduleJob, but the number of retries for that job is not updated properly, causing it to be rescheduled continuously if it continues to fail. This results in many ec2 instances being created until the server is shut down.

Issue #273, opened 2025-02-24
Spot instances for autograding jobs


KesterTan:
We should be provisioning spot instances and handling failure of spot instances instead of more expensive ec2 vms

PR #284, opened 2025-10-27
Infinite retries bugfix, Spot instances, and some code cleanup

anthony-yip:
Fixes #123, #279, #277, #273.

Sorry for the gigantic PR but all the little fixes built off of each other. 

Summary of changes:
- Preliminary mypy support and corresponding type annotations
    - Implementation of VMMS and Local/Remote as Protocols
- Code cleanup (removing dead code) in which a job always has a preVM (provided by the jobManager)
- Functional semantics for local/remote data structures (since Redis is weird)
- Fixing infinite retries (coherence bug in makeDead)
    - Immutable Attribtues for TangoJob for better maintenance
- Spot instances added
    - Errors at all stages of the pipeline (e.g. copy in) now trigger a rescheduling (instead of just failing). This is because the spot termination can occur an an arbitrary stage
    - Enforcing the invariant that detachVM is always called upon a a worker's termination

Also associated with https://github.com/autolab/Autolab/pull/2301


KesterTan (changes requested):
Can we resolve the merge conflicts

KesterTan (changes requested):
Spot instances functionality works fine, so does infinite retries, few things:

1. Seems like job retries actually mean -> 1 + retries (I think it is more common practice to have it as total number of retries) -> https://ux.stackexchange.com/questions/95984/how-to-count-retries-should-i-count-how-many-has-been-completed-or-how-many-has
2. Seems like initializeVM does not trigger a job retry? Is this desired?

anthony-yip:
initializeVM is removed entirely (it was dead code)

anthony-yip:
Moved initializeVM to the job manager rather than the worker

jobManager.py
```diff
@@ -19,22 +19,25 @@
...
+from jobQueue import JobQueue
+from tangoObjects import TangoJob, TangoQueue, TangoMachine
+from typing import List, Tuple
 from worker import Worker
 
 
+
 class JobManager(object):
-    def __init__(self, queue):
+    def __init__(self, queue: JobQueue):
```
KesterTan, line 30: Nice

requirements.txt
```diff
@@ -1,12 +1,47 @@
+async-timeout==5.0.1
```
KesterTan, line 1: Where is this being used?

requirements.txt
```diff
@@ -1,12 +1,47 @@
...
+python-dateutil==2.9.0.post0
+pytz==2025.2
+PyYAML==6.0.2
 redis==4.4.4
 requests==2.31.0
-# needed for tashi, rpyc==4.1.4
+s3transfer==0.11.5
+six==1.17.0
+supervisor==4.1.0
+tomli==2.2.1
```
KesterTan, line 31: same here

tangoObjects.py
```diff
@@ -230,16 +348,35 @@ def get(self):
...
+        ...
+    def put(self, item) -> None:
+        ...
+    def get(self, block=True, timeout=None) -> Optional[T]:
+        ...
+    def get_nowait(self) -> Optional[T]:
+        ...
+    def remove(self, item) -> None:
+        ...
+    def _clean(self) -> None:
```
KesterTan: Nit: seems to be called externally in job Queue, maybe we should not underscore it

tangoObjects.py
```diff
@@ -230,16 +348,35 @@ def get(self):
...
+        ...
+    def make_empty(self) -> None:
```
KesterTan, line 374: What is the difference between make_empty and clean? They seem to be doing the same thing

worker.py
```diff
@@ -371,35 +394,38 @@ def run(self):
...
+            #         msg = "Error: Autodriver usage error (status=%d)" % (ret["runjob"])
+            #     elif ret["runjob"] == 2:
+            #         msg = "Error: Job timed out after %d seconds" % (self.job.timeout)
+            #     elif ret["runjob"] == 3:  # EXIT_OSERROR in Autodriver
+            #         # Abnormal job termination (Autodriver encountered an OS
+            #         # error).  Assume that the VM is damaged. Destroy this VM
+            #         # and do not retry the job since the job may have damaged
+            #         # the VM.
+            #         msg = "Error: OS error while running job on VM"
+            #         detachMethod = DetachMethod.DESTROY_WITHOUT_REPLACEMENT
```
KesterTan: why not just delete these lines?

requirements.txt
anthony-yip: dependency of redis

requirements.txt
anthony-yip: dependency of boto3

tangoObjects.py
anthony-yip: fixed

tangoObjects.py
anthony-yip: changed to just make_empty

worker.py
anthony-yip: deleted

anthony-yip:
> Spot instances functionality works fine, so does infinite retries, few things:
> 
>     1. Seems like job retries actually mean -> 1 + retries (I think it is more common practice to have it as total number of retries) -> https://ux.stackexchange.com/questions/95984/how-to-count-retries-should-i-count-how-many-has-been-completed-or-how-many-has

I think it makes sense as it is, so when Config.RETRIES is 0, it attempts it once and then doesn't retry it.
> 
>     2. Seems like initializeVM does not trigger a job retry? Is this desired?

KesterTan (approved):
LGTM, nice work

PR #286, opened 2025-12-02
Type Annotations

### LINKED ISSUE #185: Type annotations

anthony-yip:
Fixes #185 .

You can now just run `mypy .` in /opt/TangoService/Tango to check for any type errors.

Because this PR includes a lint fix as the first commit, its easiest to check the diff from the second commit onwards

Changes
- Added an interface for all VMMS to conform to
- Note that tashiVMMS and distDocker are currently unsupported.


jhs-panda:
> You can now just run `mypy .` in /opt/TangoService/Tango to check for any type errors.

Nit: Should it be just in the Tango directory, instead of /opt/TangoService/Tango?

PR #287, opened 2026-01-12
Fix Dockerfile

KesterTan:
Dockerfile install docker command is outdated and causes issues when running the autograder. This PR fixes the issues by installing docker using the package manager.


coder6583 (approved):
lgtm

PR #290, opened 2026-04-16
Fixed Polling issue in frontend for ECR Builder

Giabbi:
Changes proposed in this PR:
- Refactored `ecrBuilder.py` logging from in-memory dictionaries to stateless, file-based logging for durability and thread safety.
- Updated `get_build_status` to return the complete log history on every poll, removing the brittle "consume-on-read" bug that caused fragmented logs.
- Enforced AWS ECR repository naming constraints by automatically sanitizing `course_id` and `image_name` to lowercase before executing boto3 API calls.


coder6583 (approved):
LGTM

PR #291, opened 2026-04-17
Cleanup of Local Docker Images

Giabbi:
Changes proposed in this PR:
- Added a `finally` block to the `_build_and_push_task` thread in `ecrBuilder.py` to guarantee local Docker cleanup executes regardless of whether the build succeeds or fails.
- Implemented forceful deletion (`force=True`) of the final compiled ECR image and any pulled base templates from the local EC2 instance to prevent long-term storage bloat.
- Added an automatic garbage collection step using `docker_client.images.prune(filters={'dangling': True})` to instantly remove orphaned intermediate cache layers, actively reclaiming disk space after every build run.

### DIFF
--- a/vmms/ecrBuilder.py
+++ b/vmms/ecrBuilder.py
@@ -173,6 +173,27 @@ Pin-Priority: -1
             status_id=255,
             status_msg=f"Build failed: {str(e)}",
         )
+    finally:
+        print("Cleaning up local Docker storage...")
+        try:
+            # Remove the final built image from the EC2 drive
+            if 'full_image_name' in locals():
+                docker_client.images.remove(image=full_image_name, force=True)
+            
+            # Remove the base template image (if it was pulled)
+            if base_uri is not None:
+                docker_client.images.remove(image=base_uri, force=True)
+            if base_tag is not None:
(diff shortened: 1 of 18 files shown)


vmms/ecrBuilder.py
```diff
@@ -173,6 +173,27 @@ def _build_and_push_task(job_id, course_id, image_name, dockerfile_content, base
...
         )
+    finally:
+        print("Cleaning up local Docker storage...")
+        try:
+            # Remove the final built image from the EC2 drive
+            if 'full_image_name' in locals():
+                docker_client.images.remove(image=full_image_name, force=True)
+            
+            # Remove the base template image (if it was pulled)
+            if base_uri is not None:
```
coder6583: now that i thought about it, we probably shouldn't remove base docker images. If there are multiple concurrent images building on the same shared base image, if one finishes and cleans up the base image, the other might not work.

coder6583 (changes requested):

vmms/ecrBuilder.py
Giabbi: Oh that's true! Removed those from clenup!

PR #292, opened 2026-04-17
ECR Builder / New Ec2Docker VMMS

coder6583:
## Description
This feature allows instructors to upload Dockerfiles, which will then be used to build the Docker image locally in Tango, and pushed to AWS ECR, so the autograding instances can pull the Docker image and use it to run the autograder.
Another feature we have is the public template image (base docker image). then instructors can upload Dockerfiles that pull from docker images (FROM <image name> at the top of the file), by sending the base image URI with the addJob request.
This feature also implements a new VMMS "ec2Docker" that needs to be set in config.py. It pulls a docker image from AWS ECR,  and spawns a docker container based on that image in the EC2 instance.

## Motivation and Context
Currently, there are no ways to create autograding images that can be used on ec2 instances from the frontend. 

## How Has This Been Tested?
On the Autolab side:
1. Create a course
2. On the Manage Course page, you should be able to see a button that says "Manage autograder Images" on the left side.
3. Click on New Docker Image, enter whatever name you want, and upload a dockerfile with it.
4. Wait for the docker image to be built, you can refresh the status by either refreshing the page or clicking the refresh all button. When done, it should say "Ready". The View Logs page only works when the build has finished, so you won't be able to see anything when the docker image is building. I am currently working on fixing this.
5. Create a new assessment and an autograder, and in the autograder settings, choose the newly built image. Only choose the image when it has finished building (the image shows up as "Ready" in the Manage autograder images page).
6. You should be able to submit a file and see it autograde (hopefully).

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

Need to change config.py Config.VMMS to "ec2Docker"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,10 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN chmod 1777 /tmp
 
 # Install dependancies
+RUN rm -f /etc/apt/sources.list.d/*passenger* \
+ && apt-get update \
+ && apt-get install -y sqlite3 tzdata shared-mime-info \
+ && rm -rf /var/lib/apt/lists/*
 RUN apt-get update && apt-get install -y \
 	nginx \
 	curl \
@@ -38,7 +42,7 @@ RUN apt-get update && apt-get install -y \
 
 WORKDIR /opt/TangoService/Tango/
 
-# Install Docker
+# Install Docker from Docker Inc. repositories.
 RUN set -eux; \
     apt-get update; \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends ca-certificates curl gnupg; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,58 +3,67 @@ FROM ubuntu:20.04
 
 MAINTAINER Autolab Development Team "autolab-dev@andrew.cmu.edu"
 
-# Setup correct environment variable
-ENV HOME /root
+# Setup correct environment variables
+ENV HOME=/root
+ENV DEBIAN_FRONTEND=noninteractive
 
 # Change to working directory
 WORKDIR /opt
 
-# To avoid having a prompt on tzdata setup during installation
-ENV DEBIAN_FRONTEND=noninteractive
-
 RUN chmod 1777 /tmp
 
-# Install dependancies
-RUN rm -f /etc/apt/sources.list.d/*passenger* \
- && apt-get update \
- && apt-get install -y sqlite3 tzdata shared-mime-info \
- && rm -rf /var/lib/apt/lists/*
-RUN apt-get update && apt-get install -y \
-	nginx \
-	curl \
-	git \
-	vim \
-	supervisor \
-	python3 \
-	python3-pip \
-	build-essential \
-	tcl8.6 \
-	wget \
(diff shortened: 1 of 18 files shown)


vmms/ec2Docker.py
```diff
@@ -0,0 +1,592 @@
...
+    def __init__(self, accessKeyId=None, accessKey=None):
+        """log - logger for the instance
+        connection - EC2Connection object that stores the connection
+        info to the EC2 network
+        instance - Instance object that stores information about the
+        VM created
+        """
+        # do not do anything until we acquire a vm semaphore
+        Ec2Docker.acquire_vm_semaphore()
+        validated_key_path = Ec2Docker._validate_ec2_runtime_config()
```
anthony-yip: nit: swap the order of these two lines to make it consistent with ec2SSH (and to avoid leaking the semaphore)

vmms/ecrBuilder.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+    thread.daemon = True
+    thread.start()
+
+    return int(job_id)
+
+def _build_and_push_task(job_id, course_id, image_name, dockerfile_content, base_tag, base_uri):
+    course_id = course_id.lower()
+    image_name = image_name.lower()
+    if course_id is "public":
```
anthony-yip: change to == to prevent identity comparison

Dockerfile
```diff
@@ -15,6 +15,10 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN chmod 1777 /tmp
 
 # Install dependancies
+RUN rm -f /etc/apt/sources.list.d/*passenger* \
+ && apt-get update \
+ && apt-get install -y sqlite3 tzdata shared-mime-info \
+ && rm -rf /var/lib/apt/lists/*
```
anthony-yip: why rm lists and then apt-get update right after?

restful_tango/server.py
```diff
@@ -142,6 +144,63 @@ def post(self, key):
         self.tempfile.close()
         self.write(tangoREST.build(key, name, self.request.headers["imageName"]))
 
+class BuildImageHandler(tornado.web.RequestHandler):
+    def post(self, key):
+        """post - Trigger the ECR Docker build."""
+        try:
+            payload = json.loads(self.request.body.decode('utf-8'))
+            course_id = payload.get("course_id")
+            job_id = payload.get("job_id")
```
anthony-yip, line 153: job_id assigned twice for different meanings is probably not a good idea for maintainability

restful_tango/server.py
```diff
@@ -142,6 +144,63 @@ def post(self, key):
...
+            if base_tag is not None and base_uri is None:
+                self.set_status(400)
+                self.write({"statusMsg": "Requires the URI to pull from a base docker image", "statusId": -1})
+                return
+            
+            if is_public:
+                course_id = "public"
+
+            # Trigger background build
+            job_id = tangoREST.buildImage(key, course_id, job_id, image_name, dockerfile_content, base_tag, base_uri)
```
anthony-yip: see here

anthony-yip (changes requested):
.

